PR #22, opened 2026-05-18
fix: commit-msg hook blocks merge commits

chaudharydeepanshu:
Blocks local git merge — all promotions via GitHub PRs (squash-merge).

PR #23, opened 2026-05-18
ci: commit-msg hook (merge blocker + printf), promotion-check, hooksPath docs

chaudharydeepanshu:
Fresh PR from dev. Supersedes #22 which had rebase conflicts.

- commit-msg hook: blocks merge commits, allows reverts, uses printf
- promotion-check workflow: feature → dev → main only
- CONTRIBUTING.md: documents core.hooksPath setup

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -5,7 +5,18 @@
 
 msg=$(head -1 "$1")
 
-if ! echo "$msg" | grep -qE '^(feat|fix|docs|chore|ci|test|refactor|perf|build|deps)(\(.+\))?!?: .+'; then
+# Allow reverts (squash-reverts have conventional titles, but git revert doesn't)
+printf '%s\n' "$msg" | grep -qE '^Revert ' && exit 0
+
+# Block merge commits — all branch-to-branch moves go through PRs (squash-merge)
+if printf '%s\n' "$msg" | grep -qE '^Merge '; then
+  echo "ERROR: Merge commits are not allowed."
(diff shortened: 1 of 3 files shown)


slopfairy:
reviewed ✨

slopfairy (approved):
Looks clean and well-organized. Here's what stood out:

---

✅ **`.githooks/commit-msg`**
- `printf` over `echo` — good portability choice
- Revert allowance correctly placed *before* the merge block, so `Revert "Merge ..."` passes as intended
- Merge commit rejection gives a clear, actionable error message that explains *why* (squash-merge workflow) not just *what*

✅ **`.github/workflows/promotion-check.yml`**
- Minimal and focused — no unnecessary checkout step, no unused permissions
- `::error::` annotations will surface clearly in the GitHub UI
- Covers both directions: blocks non-dev → main *and* main → dev

✅ **`CONTRIBUTING.md`**
- Docs the `core.hooksPath` config inline with setup commands so new contributors won't miss it

---

🟡 **Suggestion (non-blocking):** Consider declaring `permissions: {}` in `promotion-check.yml` to make the least-privilege posture explicit. Since the job doesn't use `GITHUB_TOKEN` at all, this prevents a future maintainer from accidentally relying on default token scopes when adding steps. Not required — just future-proofing.

---

Nice work on all three fronts. The diff is compact, each change is independently useful, and the consistency between the hook logic and the promotion-check workflow shows good architectural thinking.

Verdict: APPROVE

PR #29, opened 2026-06-02
fix: publish.yml errors on branch push — replace all bare inputs.tag

chaudharydeepanshu:
## Summary
- Replace all 9 `inputs.tag` references with `github.event.inputs.tag`
- GitHub evaluates all expressions for every push event, even when the `tags: ['v*']` filter would exclude it
- Bare `inputs.tag` errors when the inputs context doesn't exist (non-dispatch events)
- `github.event.inputs.tag` returns null gracefully on non-dispatch events

## Test plan
- [ ] Push to dev no longer triggers a failed publish.yml run
- [ ] Manual workflow_dispatch with a tag still works
- [ ] Tag push still triggers the full compile + publish pipeline

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,7 +29,7 @@ env:
   COMPILE_OUTPUT_DIR: out
 
 concurrency:
-  group: release-${{ github.event.inputs.tag || github.ref_name }}
+  group: release-${{ github.event.github.event.inputs.tag || github.ref_name }}
   cancel-in-progress: false
 
 jobs:
@@ -38,7 +38,7 @@ jobs:
     runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
-        with: { ref: ${{ inputs.tag || github.ref_name }}, submodules: recursive }
+        with: { ref: ${{ github.event.inputs.tag || github.ref_name }}, submodules: recursive }
       - uses: ./.github/actions/compile-macos
       - uses: actions/upload-artifact@v4
         with: { name: bin-macos, path: out/ }
@@ -46,7 +46,7 @@ jobs:
     runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
-        with: { ref: '${{ inputs.tag || github.ref_name }}', submodules: recursive }
+        with: { ref: '${{ github.event.inputs.tag || github.ref_name }}', submodules: recursive }
       - uses: ./.github/actions/compile-ios
       - uses: actions/upload-artifact@v4
         with: { name: bin-ios, path: out/ }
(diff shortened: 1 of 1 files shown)


slopfairy:
## Review results for PR #29

**Verdict: REQUEST_CHANGES** (submitted via review)

### What was reviewed

Changes to `.github/workflows/publish.yml` — replacing bare `inputs.tag` with `github.event.inputs.tag` to fix workflow errors on branch push events.

### Findings

**🔴 Blocking:**
- **.github/workflows/publish.yml:32** — The concurrency group line was already using `github.event.inputs.tag` in the original. The global replace incorrectly changed it to `github.event.github.event.inputs.tag`, which is not a valid GitHub expression. An inline suggestion has been posted to revert this line.

**✅ Correct:**
- All 9 remaining occurrences of `inputs.tag` are correctly replaced with `github.event.inputs.tag`
- The `|| github.ref_name` fallback is preserved in all cases
- No other files modified

**Suggestion to apply the fix:**
```yaml
# Line 32, change:
  group: release-${{ github.event.github.event.inputs.tag || github.ref_name }}
# back to:
  group: release-${{ github.event.inputs.tag || github.ref_name }}
```

Once that single line is reverted, the PR is good to merge.

slopfairy (changes requested):
placeholder - will submit with final review

.github/workflows/publish.yml
```diff
@@ -29,7 +29,7 @@ env:
   COMPILE_OUTPUT_DIR: out
 
 concurrency:
-  group: release-${{ github.event.inputs.tag || github.ref_name }}
+  group: release-${{ github.event.github.event.inputs.tag || github.ref_name }}
```
slopfairy, line 32: **Bug introduced**: The concurrency group line already used the correct `github.event.inputs.tag` in the original file. This change accidentally produces `github.event.github.event.inputs.tag` which is not a valid expression.

```suggestion
  group: release-${{ github.event.inputs.tag || github.ref_name }}
```

slopfairy (changes requested):
## Verdict: REQUEST_CHANGES

The intent is correct — replacing `inputs.tag` with `github.event.inputs.tag` — but a regex/global-replace overrun introduced one bug.

### 🔴 Blocking (must fix before merge)

1. **.github/workflows/publish.yml:32** — The concurrency group line was already using `github.event.inputs.tag` in the original file (`dev`). The global replace incorrectly turned it into `github.event.github.event.inputs.tag`, which is not a valid GitHub context path. Revert this line back to `github.event.inputs.tag`. A suggestion is posted inline.

### ✅ Looks good

- All 9 occurrences of bare `inputs.tag` on push-only event paths are correctly replaced with `github.event.inputs.tag`, which gracefully evaluates to `null` on non-dispatch events.
- The `|| github.ref_name` fallback is preserved in all cases, so tag-push workflow runs still resolve correctly.
- No other files modified; the change scoped exactly to the broken expressions.

PR #35, opened 2026-06-03
fix: YAML syntax error in create-release.yml

chaudharydeepanshu:
## Summary
- Bare `<details>` HTML inside YAML `run:` block caused parse error
- Replaced with `printf` to build the string without inline HTML in YAML source
- Validated all 7 workflow files with `yaml.safe_load` — all pass

## Test plan
- [ ] create-release.yml no longer shows "Invalid workflow file" on push

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -115,13 +115,9 @@ jobs:
             fi
             if [ -n "$COMMITS" ]; then
               COMMIT_LIST=$(echo "$COMMITS" | sed 's/^/- /')
-              NOTES="$NOTES
-
-<details><summary>Commits since ${PREV_TAG:-initial} ($COUNT)</summary>
-
-$COMMIT_LIST
-
-</details>"
+              DETAILS_OPEN="<details><summary>Commits since ${PREV_TAG:-initial} ($COUNT)</summary>"
+              DETAILS_CLOSE="</details>"
+              printf -v NOTES '%s\n\n%s\n\n%s\n\n%s' "$NOTES" "$DETAILS_OPEN" "$COMMIT_LIST" "$DETAILS_CLOSE"
             fi
 
             echo "Creating release $TAG (prerelease=$IS_PRE)..."


slopfairy:
## Review complete

**Verdict: APPROVE** 

Reviewed commit `c3cb4ee` — one file changed (`.github/workflows/create-release.yml`).

### What was checked

- **Output equivalence**: The `printf -v` replacement produces identical content to the original multi-line string: `$NOTES\n\n<details>...</summary>\n\n$COMMIT_LIST\n\n</details>`. The blank-line structure matches exactly.
- **Shell safety**: `printf -v` is a bash builtin, and the script already uses bash-specific features (`[[ ]]`, etc.). No compatibility concern on the GitHub Actions Ubuntu runner.
- **No format-string injection**: All variables are passed as `%s` arguments, so special characters (including `%`) in `$NOTES` or `$COMMIT_LIST` are safe.
- **Consistency**: Pattern matches `publish.yml` line 205, which already uses `printf` with `<details>` in single quotes for the same purpose.
- **YAML validity**: Avoids bare HTML tags in the literal block scalar, which was the reported parse error trigger.

No issues found.

slopfairy (approved):
## Verdict: APPROVE

Fix is technically correct and consistent with the existing codebase.

### ✅ What was checked

- **Bash correctness**: `printf -v` produces identical output to the original multi-line string — `$NOTES`, `$DETAILS_OPEN`, `$COMMIT_LIST`, and `$DETAILS_CLOSE` are concatenated with the same `\n\n` separators. No trailing newline difference either.
- **Shell compatibility**: `printf -v` is a bash builtin; the script already uses bash features (`[[ ]]`, process substitution patterns), and the GitHub Actions runner defaults to bash. No regression.
- **Consistency**: `publish.yml` line 205 already uses `printf \n<details>...` for the same `<details> block pattern. This PR brings `create-release.yml` in line.
- **Security**: No user-controlled data in format strings — all variables are passed as `%s` arguments, so no format string injection.
- **YAML validity**: The fix avoids bare HTML tags in the YAML literal block scalar, eliminating the parse error trigger. Validated against `yaml.safe_load` per PR description.

PR #37, opened 2026-06-03
fix: API-based asset hashes, MSVC-first Windows compile

chaudharydeepanshu:
## Summary
- `write_asset_hashes.dart` queries GitHub API for SHA256 digests instead of downloading 600+ MB of binaries
- `compile_rust.sh windows` prefers MSVC over MinGW on Windows runners — gets arm64 too
- All 7 workflow YAML files validated

## Test plan
- [ ] Delete release + tag, merge, retrigger full flow
- [ ] publish.yml generates hashes via API (instant, no downloads)
- [ ] Windows compile produces both x64 + arm64


slopfairy:
## Review Complete — PR #37

**Verdict: APPROVE**

Three clean files, no blocking issues.

### Files changed

| File | Change |
|------|--------|
| `.github/workflows/publish.yml` | Added `GH_TOKEN` env to hash step + ephemeral git commit before `pub publish` |
| `tool/compile_rust.sh` | MSVC-first on Windows (x64 + arm64), MinGW cross-compile fallback |
| `tool/write_asset_hashes.dart` | Rewrote from binary download (~600 MB) to GitHub API SHA256 digest lookup |

### Reviewed and confirmed

- Asset naming alignment: release assets are named `<platform>-<libname>` (traced through flatten step → `compile_one` output). The API-based tool reads these same names, so `asset_hashes.dart` keys match `hook/build.dart` consumption.
- Graceful degradation for missing digests (SKIP → "no hash to verify" mode).
- Ephemeral commit is the correct fix for `dart pub publish` requiring a clean tree.
- MSVC priority swap is correct: Windows native builds get x64 + arm64, MinGW cross-compile stays as fallback.
- Workflow security is sound: no `pull_request_target`, auto-scoped `GITHUB_TOKEN`, tag-ref checkout.
- `crypto` dep kept — still used by `hook/build.dart` for SHA256 verification.

### Non-blocking note

The per-line `jsonDecode` approach in `write_asset_hashes.dart` works for current `gh api --jq` output format. Consider `--slurp` + single `jsonDecode` for future-proofing. Also consider reading repo from `GITHUB_REPOSITORY` env var for fork-friendliness.

slopfairy (approved):
## Verdict: APPROVE

Three clean files, no blocking issues.

1. **tool/write_asset_hashes.dart:29** — The per-line `jsonDecode` approach works for `gh api --jq` output, but it assumes the release endpoint's `--jq` filter always emits one object per line. If the API response changes or assets are restructured in a future `gh` version, the line-splitting could silently produce partial output. Consider wrapping the entire `gh api` response in `--slurp` to emit a single JSON array, then decode once with `jsonDecode`. Not blocking for this PR since the current format is stable.

2. **tool/write_asset_hashes.dart** — The hardcoded `_repo = 'whuppi/pdf_manipulator'` makes this tool fork-specific. If this repo is ever forked for development, the tool would point at the upstream release instead of the fork's. Consider reading the repository from `GITHUB_REPOSITORY` env var (available in GHA) and falling back to the current constant for manual runs. Low priority — the publish workflow only runs on the canonical repo.

### ✅ Looks good

- **Asset key naming alignment**: Verified that release assets are named `<platform>-<libname>` (e.g., `macos-arm64-libpdf_oxide.dylib`) by tracing the flatten step in `publish.yml` through to the `compile_rust.sh` `compile_one` output paths. The new API-based approach reads these same names, so the generated `asset_hashes.dart` keys match what `hook/build.dart` expects in `_downloadPrebuilt`. No mismatch.

- **Graceful degradation**: Missing or null digests are handled with a SKIP message, and the build hook already supports "no hash to verify" mode. You won't crash on releases without digest metadata.

- **Ephemeral commit**: `dart pub publish` requires a clean working tree. After stamping version files and generating hashes, `git add -A && git commit` is the correct fix. `--allow-empty` handles the edge case where nothing changed.

- **MSVC-first Windows**: Correct priority swap. On real Windows (MINGW/MSYS), MSVC compiles both x64 and arm64. MinGW cross-compile from Linux stays as a fallback for Linux-based Windows builds. The `do_windows` call in `do_native` on MSYS/MINGW hosts is consistent with the new branch order.

- **Workflow security**: `pull_request_target` is not used. The `GH_TOKEN` is `secrets.GITHUB_TOKEN` (auto-scoped, not a PAT). Checkout uses a tag ref, not a PR head. No injection vectors from the `${{ needs.validate.outputs.tag }}` interpolation (it's validated with a regex in the `validate` job).

- **No orphaned dependencies**: `crypto` is still imported by `hook/build.dart` for SHA256 verification — removing it from `write_asset_hashes.dart` is correct and doesn't break consumers.

PR #41, opened 2026-06-03
fix: move asset_hashes.dart to lib/src/hook/

chaudharydeepanshu:
pub.dev only allows build.dart and link.dart in hook/. Moved to lib/src/hook/ following sqlite3.dart's pattern.


slopfairy:
## Verdict: APPROVE

Three clean file changes. No blocking issues.

### ✅ Looks good

**Changes reviewed:**
1. **`hook/asset_hashers.dart` → `lib/src/hook/asset_hashers.dart`** — Renamed to comply with pub.dev's restriction that only `build.dart` and `link.dart` may live in `hook/`. This follows the same pattern used by `sqlite3.dart`. The comment was also updated to correctly reference `stamp_release.sh` instead of the old `write_asset_hashers.dart`.

2. **`hook/build.dart:26`** — Import updated from relative (`import 'asset_hashes.dart'`) to package-absolute (`import 'package:pdf_manipulator/src/hook/asset_hashes.dart'`), as required since the file is no longer in the same directory.

3. **`tool/stamp_release.sh:69`** — Path updated from `hook/asset_hashes.dart` to `lib/src/hook/asset_hashes.dart` to stay consistent with the move.

1. **`SECURITY.md`** and **`docs/ARCHITECTURE.md`** still reference the old path `hook/asset_hashes.dart`. These docs should be updated to `lib/src/hook/asset_hashes.dart` in a follow-up PR to keep documentation accurate.

slopfairy (dismissed):

slopfairy:
This PR was merged by @chaudharydeepanshu at 08:10 UTC on 2026-06-03. No further action needed.

The synchronize event arrived after the merge, so this is a no-op. The changes (moving `asset_hashes.dart` to `lib/src/hook/`, updating `ARCHITECTURE.md`, and fixing the import/stamp paths) were already applied and merged successfully.

PR #45, opened 2026-06-03
fix: push stamped commit before creating release tag

chaudharydeepanshu:
gh release create --target SHA fails when the SHA only exists locally. Push to a temp branch first, create the release, then delete the branch.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -111,8 +111,12 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git add -A
-          STAMPED_SHA=$(git commit-tree -p HEAD -m "release: $TAG" $(git write-tree))
+          git commit -m "release: $TAG"
+          STAMPED_SHA=$(git rev-parse HEAD)
           echo "Stamped commit: $STAMPED_SHA"
+
+          # Push the stamped commit so GitHub knows about it before we create the tag
+          git push origin "$STAMPED_SHA:refs/heads/_release-staging"
 
           # Generate release notes and create release at the stamped commit
           NOTES=$(bash tool/stamp_release.sh "$TAG" --github-notes)
@@ -124,6 +128,9 @@ jobs:
             --notes "$NOTES" \
             $FLAGS
           echo "Created release $TAG at $STAMPED_SHA"
+
+          # Clean up the staging branch (tag holds the reference now)
+          git push origin --delete refs/heads/_release-staging 2>/dev/null || true
 
           echo "tag=$TAG" >> "$GITHUB_OUTPUT"
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"


slopfairy:
## Review submitted for PR #45

**PR:** `fix: push stamped commit before creating release tag` by @chaudharydeepanshu

**Verdict:** REQUEST_CHANGES

**Summary:**
The PR correctly addresses the root cause — `gh release create --target SHA` fails when the SHA only exists locally. Pushing the stamped commit to a temporary staging branch before creating the release, then cleaning it up, solves this properly. The migration from `git commit-tree` (detached commit) to `git commit` (branch commit) is a natural simplification.

One blocking issue was found:

1. **Force-push needed for staging branch** — With `concurrency.cancel-in-progress: true`, a cancelled run can leave a stale `_release-staging` branch. The next run's `git push origin SHA:refs/heads/_release-staging` would fail because the existing branch points to a different SHA. Using `--force` is the right fix since the branch is ephemeral by design.

Two non-blocking items: stale `git commit-tree` references in `stamp_release.sh` comments (lines 12, 35, 60).

Inline suggestion was posted on line 119 of the workflow file.

slopfairy (changes requested):
Initial review of the fix for push-before-release workflow.

.github/workflows/create-release.yml
```diff
@@ -111,9 +111,13 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git add -A
-          STAMPED_SHA=$(git commit-tree -p HEAD -m "release: $TAG" $(git write-tree))
+          git commit -m "release: $TAG"
+          STAMPED_SHA=$(git rev-parse HEAD)
           echo "Stamped commit: $STAMPED_SHA"
 
+          # Push the stamped commit so GitHub knows about it before we create the tag
+          git push origin "$STAMPED_SHA:refs/heads/_release-staging"
```
slopfairy, line 119: This push can fail when a previous run was cancelled mid-execution (this workflow has `concurrency.cancel-in-progress: true`). If cancellation happens between this push and the cleanup on line 133, a stale `_release-staging` branch remains. The next run's push is rejected because the branch already exists pointing at a different commit.

**Fix:** use force-push:
```suggestion
          git push origin --force "$STAMPED_SHA:refs/heads/_release-staging"
```

slopfairy (changes requested):
## Verdict: REQUEST_CHANGES

Good fix. The core approach — pushing the stamped commit to a temp branch so `gh release create --target` can resolve it — is correct and solves the original problem.

### 🔴 Blocking (must fix before merge)

1. **.github/workflows/create-release.yml:119** — `git push origin "$STAMPED_SHA:refs/heads/_release-staging"` fails when a previous run was cancelled before reaching the cleanup step at line 133 (this workflow has `concurrency.cancel-in-progress: true`). The existing `_release-staging` branch points to a different SHA, so the non-force push is rejected. Use `--force` since this branch is ephemeral: `git push origin --force "$STAMPED_SHA:refs/heads/_release-staging"`

### 🟡 Non-blocking (follow-up)

1. **tool/stamp_release.sh:12** — Comment still reads "Called by CI discover job BEFORE git commit-tree" but the workflow now uses `git commit`. Similarly lines 35 and 60 reference `git commit-tree`. Worth updating to avoid confusion for future readers.

### ✅ Looks good

- The migration from `git commit-tree` (detached commit) to `git commit` (normal branch commit) is correct — the temp branch push ensures GitHub knows the SHA.
- The cleanup step properly tolerates failure (`|| true`) so a missing staging branch doesn't break the workflow.
- `contents: write` permission is sufficient for push/delete on non-protected branches.
- Concurrency group scoping (per branch) means overlapping dev/prod runs won't interfere with each other's staging branches.

PR #47, opened 2026-06-03
fix: auto-detect wasm-bindgen-cli version from Cargo.lock

chaudharydeepanshu:
Reads from Cargo.lock, installs if mismatched. Single source of truth. No hardcoded versions.

### DIFF
--- a/.github/actions/compile-wasm/action.yml
+++ b/.github/actions/compile-wasm/action.yml
@@ -7,9 +7,7 @@ runs:
     - uses: ./.github/actions/setup-rust
       with: { targets: wasm32-unknown-unknown }
     - shell: bash
-      run: |
-        cargo install wasm-bindgen-cli --version 0.2.121
-        sudo apt-get update && sudo apt-get install -y binaryen
+      run: sudo apt-get update && sudo apt-get install -y binaryen
     - shell: bash
       run: bash tool/compile_rust.sh wasm
     - shell: bash
--- a/.github/actions/setup-web-testing/action.yml
+++ b/.github/actions/setup-web-testing/action.yml
@@ -62,7 +62,6 @@ runs:
     # ── WASM toolchain + build ──
     - shell: bash
       run: |
-        cargo install wasm-bindgen-cli --version 0.2.121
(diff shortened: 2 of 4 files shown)


slopfairy:
## Analysis

PR #47 by @chaudharydeepanshu is already **merged** (merged by the author at `2026-06-03T11:45:46Z`, 14 seconds after creation).

### Changes reviewed retrospectively

The PR removes hardcoded `wasm-bindgen-cli --version 0.2.121` from two CI action files and adds auto-detection logic in `tool/compile_rust.sh:do_wasm()` that reads the version from `vendor/pdf_oxide/Cargo.lock` before each WASM build.

**4 files changed, +14/−5:**
- `.github/actions/compile-wasm/action.yml` — removed hardcoded `cargo install wasm-bindgen-cli --version 0.2.121`
- `.github/actions/setup-web-testing/action.yml` — removed hardcoded `cargo install wasm-bindgen-cli --version 0.2.121`
- `docs/UPDATING.md` — updated docs to reflect auto-detection approach
- `tool/compile_rust.sh:208-214` — version extraction + conditional install logic

### Assessment

**Design:**
- Single source of truth approach is correct — version is now derived from Cargo.lock rather than hardcoded in two separate CI files.
- Conditional install (`if [[ "$wb_installed" != "$wb_required" ]]`) avoids unnecessary re-installation on repeated runs.
- Docs update accurately describes the new behavior.

**One note for future robustness (non-blocking):** The version extraction pipeline at line 208 runs under `set -euo pipefail`. If `Cargo.lock` does not exist or does not contain `wasm-bindgen` (e.g., on a fresh checkout before the first `cargo build`), the script will exit with a non-zero code rather than a descriptive message. The affected pipeline is:

```bash
wb_required=$(grep -A1 'name = "wasm-bindgen"' "$VENDOR/Cargo.lock" | grep version | head -1 | sed 's/.*"\(.*\)"/\1/')
```

For CI this is acceptable (failure is fail-fast), but could be made more resilient with `|| wb_required=""` after the command substitution.

**Verdict:** Clean, well-scoped change. Already merged.

PR #48, opened 2026-06-03
fix: force-add Cargo.lock in stamp-tag (gitignored in vendor)

chaudharydeepanshu:
vendor/.gitignore excludes Cargo.lock. Use git add --force so the stamped tag commit includes it for deterministic builds + wasm-bindgen version detection.

### DIFF
--- a/tool/stamp_release.sh
+++ b/tool/stamp_release.sh
@@ -82,7 +82,10 @@ if [[ "$MODE" == "--stamp-tag" ]]; then
       # Remove submodule registration so git tracks files directly
       git rm --cached "$sub" 2>/dev/null || true
       rm -rf "$sub/.git"
-      git add "$sub/"
+      # Force-add everything including files excluded by vendor .gitignore
+      # (Cargo.lock is gitignored in library crates but we need it for
+      # deterministic builds + wasm-bindgen-cli version detection)
+      git add --force "$sub/"
       echo "  $sub → raw source (de-registered submodule)"
     fi
   done


slopfairy:
## Analysis — PR #48: force-add Cargo.lock in stamp-tag

**Status:** Already merged by @chaudharydeepanshu. No action needed.

### What was checked

Confirmed the rationale by examining the submodule chain:

- `vendor/pdf_oxide` is a submodule pointing to `whuppi/pdf_oxide`
- `pdf_oxide/.gitignore` contains an **active** `/Cargo.lock` entry (the standard Rust convention for library crates), which means `Cargo.lock` is gitignored inside the submodule
- The stamp-tag workflow in `tool/stamp_release.sh` de-registers submodules (`git rm --cached` + `rm -rf .git/`) and then adds the vendored source directly into the release tag commit
- Without `--force`, `git add "$sub/"` respects the submodule's `.gitignore`, which still exists in the working tree, and silently skips `Cargo.lock`

### Verdict: correct

The change from `git add "$sub/"` to `git add --force "$sub/"` (line 88) is the right fix:
- `--force` overrides `.gitignore` exclusions without affecting other behavior
- The release tag commit should contain `Cargo.lock` for deterministic builds and wasm-bindgen-cli version detection
- No risk of unwanted artifacts: the stamp-tag runs in CI from a fresh checkout where generated build artifacts (e.g. `target/`, `pkg/`) do not exist, and `target/` is also covered by the root `.gitignore`
- The inline comments (lines 85-87) accurately document the rationale

Change is minimal (1 line + comments), correct, and well-documented.

PR #49, opened 2026-06-03
fix: add vendor false_secrets at stamp-tag time

chaudharydeepanshu:
Vendor test PEM keys are only shipped in the stamped tag commit. false_secrets /vendor/** added by stamp_release.sh --stamp-tag, not permanently in pubspec.

### DIFF
--- a/tool/stamp_release.sh
+++ b/tool/stamp_release.sh
@@ -94,6 +94,12 @@ if [[ "$MODE" == "--stamp-tag" ]]; then
     echo "  .gitmodules removed"
   fi
 
+  # Add false_secrets for vendor test keys (only needed when vendor is shipped)
+  if ! grep -q '/vendor/\*\*' pubspec.yaml; then
+    sed -i.bak '/^false_secrets:/a\  - /vendor/**' pubspec.yaml && rm -f pubspec.yaml.bak
+    echo "  pubspec.yaml += false_secrets /vendor/**"
+  fi
+
   echo "=== Tag tree ready ==="
   exit 0
 fi


slopfairy:
## PR #49 Review — `fix: add vendor false_secrets at stamp-tag time`

**Status**: Already merged by @chaudharydeepanshu.

**Change**: Adds 6 lines to `tool/stamp_release.sh` (line 97-101) in the `--stamp-tag` block to conditionally append `  - /vendor/**` under `false_secrets:` in `pubspec.yaml` when preparing the release tag tree.

### Assessment

The change is correct and well-structured:

- **Scoped correctly**: Only runs during `--stamp-tag` mode, when vendor submodules have been de-registered and raw source files (including test PEM keys) are shipped in the tag commit. The default `--stamp` mode for pub.dev publish does not include vendor, so no exclusion is needed there.
- **Idempotent**: Guards with `grep -q '/vendor/\*\*'` to avoid duplicate entries if the script runs multiple times.
- **Consistent style**: Uses the same `sed -i.bak ... && rm -f ...bak` pattern used elsewhere in the script (lines 71, 74) for in-place edits with cleanup.
- **Valid YAML**: The current `false_secrets:` section in `pubspec.yaml` is a multi-line YAML list; appending another `- /vendor/**` entry is syntactically correct.
- **Path matches reality**: `/vendor/**` correctly matches the de-registered vendor directory that now contains the test PEM key files.
- **Edge cases**: If `false_secrets:` key is removed from pubspec.yaml in the future, the `sed` pattern simply won't match; exit code 0, no harm (the `.bak` file is cleaned up). If the vendor keys are already in pubspec, the `grep` guard skips.

No issues found. This is a focused, correct change that keeps false_secret exclusions out of the dev branch while preventing secret-scanner noise in the release tag.

**Verdict**: APPROVED (post-merge confirmation — change is sound).

PR #67, opened 2026-06-05
fix(ci): version-level release concurrency + branch-file gate

chaudharydeepanshu:
## Summary

- Replaced workflow-level `concurrency` (branch-keyed, `cancel-in-progress`) with a gate job + version-level concurrency on the discover job
- Gate job checks via `git diff` whether the branch's relevant changelog file actually changed (dev→`CHANGELOG.pre.md`, prod→`CHANGELOG.md`). Falls back to `should_run=true` when diff fails (force-push, workflow_dispatch)
- Concurrency group is now `release-{version}` — same version cancels stale run, different versions run independently
- Fixes: CHANGELOG.md push to dev no longer cancels an in-progress prerelease, and no longer triggers a useless stable-on-dev run

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -30,21 +30,66 @@ permissions:
 env:
   COMPILE_OUTPUT_DIR: out
 
-concurrency:
-  # github.event.inputs.branch (not bare inputs.branch) — top-level expressions
-  # are evaluated for ALL push events, not just workflow_dispatch. Bare inputs.*
-  # errors when the inputs context doesn't exist on push events.
-  group: release-${{ github.event.inputs.branch || github.ref_name }}
-  cancel-in-progress: true
-
 jobs:
+  # ── Gate — should this push trigger a release? ─────────────────────
+  # dev only reacts to CHANGELOG.pre.md changes, prod to CHANGELOG.md.
+  # Uses git diff across the full push range (not head_commit.modified,
+  # which misses force-pushes and ff-merges). Falls back to should_run=true
+  # when diff fails (new branch, orphaned before-SHA) — a wasted run is
+  # cheap; a missed release is not.
+  gate:
+    name: Release gate
+    runs-on: ubuntu-latest
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
+      version: ${{ steps.check.outputs.version }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.branch || github.ref_name }}
+
+      - name: Check changelog relevance
+        id: check
+        env:
+          BRANCH: ${{ inputs.branch || github.ref_name }}
+          BEFORE: ${{ github.event.before }}
+          AFTER: ${{ github.sha }}
+        run: |
+          if [[ "$BRANCH" == "prod" ]]; then
+            TARGET_FILE="CHANGELOG.md"
+          else
+            TARGET_FILE="CHANGELOG.pre.md"
+          fi
+
+          CHANGED_FILES=$(git diff --name-only "$BEFORE" "$AFTER" 2>/dev/null || echo "$TARGET_FILE")
+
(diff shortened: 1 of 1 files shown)


slopfairy:
## Review complete — APPROVED

I reviewed PR #67 by @chaudharydeepanshu.

### What was changed
- **Removed** the top-level `concurrency` block (branch-keyed, cancelled in-progress across all runs on the same branch)
- **Added** a `gate` job that runs `git diff` to check whether the relevant changelog file actually changed in the push (dev→`CHANGELOG.pre.md`, prod→`CHANGELOG.md`), with safe fallback-to-run on diff failure
- **Moved** concurrency to the `discover` job at version-level (`release-{version}`), so same-version pushes cancel stale runs but different versions proceed independently

### Verdict: APPROVE

Key findings:
- The gate job design is sound: safe fallback on failure, correct branch-to-file mapping, concise implementation
- Version-level concurrency correctly fixes the stated bug (CHANGELOG.md push to dev no longer cancels a prod prerelease and no longer triggers a useless run)
- The `workflow_dispatch` edge case is handled by the fallback (always runs on manual dispatch, which is the desired behavior)
- No security concerns — no new secrets exposure, no `pull_request_target` usage, same `contents: write` scope as before

Non-blocking observation: the `github.event.before` → empty-string behavior on `workflow_dispatch` is subtle and could be explicitly documented in the gate's block comment, though the fallback handles it correctly.

### Review body
Submitted inline with full details.

slopfairy (dismissed):
---
## Verdict: APPROVE

Well-structured change. The version-level concurrency is the right fix for the race condition described, and the gate job is a clean approach to avoid spurious triggers. The `2>/dev/null || echo` fallback properly handles the edge cases where `github.event.before` is unavailable (new branch, `workflow_dispatch`).

1. **`.github/workflows/create-release.yml:57`** — `github.event.before` is empty on `workflow_dispatch` events. The `git diff` will fail (as expected) and the fallback sets `should_run=true`, which is correct. Worth being explicit about this in the block comment above the gate job since it's a subtle behavior: maybe note that the empty-string `BEFORE` causes an intentional diff failure that triggers the safe fallback. (Trivial documentation nit, not a code issue.)

2. **`.github/workflows/create-release.yml:91`** — If the version string can't be parsed from the changelog (e.g., malformed heading), the concurrency group falls back to `release-unknown` for all such runs. Since the `discover` job would also fail to extract a version and set `has_release=false`, this only affects the `discover` job's own cancellation behavior — the downstream compile jobs wouldn't run anyway. No action needed, just an observation.

### ✅ Looks good
- The gate-to-discover dependency chain is logically sound
- Branch/file mapping (dev → CHANGELOG.pre.md, prod → CHANGELOG.md) is consistent with the discover job's own logic
- Fallback-on-failure for `git diff` is the correct tradeoff (wasted run vs missed release)
- `cancel-in-progress` at version-level correctly allows independent versions to proceed in parallel

slopfairy:
## Re-review complete — APPROVED (re-approval after push)

PR #67 by @chaudharydeepanshu received a new commit: `be6a567` — "docs(ci): clarify workflow_dispatch fallback in gate comment".

### What changed

The only change is in the gate job's block comment. The original read:
```
# Falls back to should_run=true when diff fails
```

The updated comment now explicitly lists the edge cases:
```
# Falls back to should_run=true when diff fails (new branch, orphaned
# before-SHA, or workflow_dispatch where github.event.before is empty)
```

This directly addresses the non-blocking observation from my first review about documenting the `workflow_dispatch` → `github.event.before === ""` behavior. No functional changes to the workflow logic.

### Verdict: APPROVE

No new issues introduced. The comment clarification is a net improvement.

### ✅ Looks good

- Gate job logic is unchanged and correct
- Concurrency model (version-level on discover job) is unchanged
- Comment now explicitly documents the `workflow_dispatch` edge case for future readers

slopfairy (approved):
---
## Verdict: APPROVE

Re-review of new commit `be6a567` — a documentation-only change that addresses the non-blocking observation from my first review. No functional changes. Re-approving.

PR #85, opened 2026-06-09
fix(ci): release gate only triggers on new changelog versions

chaudharydeepanshu:
## Summary

- Release pipeline was triggering on any `CHANGELOG.md` / `CHANGELOG.pre.md` edit — typos, formatting, reordering existing entries all triggered a full release build.
- Gate now compares version headers before vs after the push using `comm -23`. Only triggers when a NEW `## version` header exists that wasn't in the previous commit.

## Test plan

- [ ] Edit existing changelog entry, push to dev — gate skips
- [ ] Add new `## 1.0.6-dev.0` header, push to dev — gate triggers

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -14,18 +14,27 @@ permissions:
   pull-requests: write
 
 env:
-  RUNNER: ubuntu-24.04
+  DEFAULT_RUNNER: ubuntu-24.04
 
 jobs:
+  inputs:
+    runs-on: ubuntu-24.04
+    outputs:
+      runner: ${{ steps.resolve.outputs.runner }}
+    steps:
+      - id: resolve
+        run: echo "runner=${{ env.DEFAULT_RUNNER }}" >> $GITHUB_OUTPUT
+
   # Revoke stale ready-to-test label when new commits are pushed.
   # Lives here (not full-test.yml) because pull_request_target gives
   # write access to labels on fork PRs.
   revoke-ready-to-test:
+    needs: inputs
     if: >-
       github.event_name == 'pull_request_target' &&
       github.event.action == 'synchronize' &&
       contains(github.event.pull_request.labels.*.name, 'ready-to-test')
-    runs-on: ${{ env.RUNNER }}
+    runs-on: ${{ needs.inputs.outputs.runner }}
     steps:
       - name: Remove label + notify
         env:
@@ -43,12 +52,13 @@ jobs:
   # run dependabot commands (no push access), so we comment asking a human.
   # Skips when dependabot itself pushes (after a recreate) to avoid loops.
   dependabot-needs-recreate:
+    needs: inputs
     if: >-
       github.event_name == 'pull_request_target' &&
       github.event.action == 'synchronize' &&
       github.event.pull_request.user.login == 'dependabot[bot]' &&
       github.event.sender.login != 'dependabot[bot]'
-    runs-on: ${{ env.RUNNER }}
+    runs-on: ${{ needs.inputs.outputs.runner }}
     steps:
       - name: Notify maintainer
         env:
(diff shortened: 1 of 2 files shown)


slopfairy:
## Review complete — **APPROVED**

Reviewed `tool/release.sh` — the `cmd_gate()` function change in PR #85.

### Summary

The change replaces the old behavior (trigger on any changelog edit) with version-header-aware gating. It uses `comm -23` to compare the set of `## version` headers before vs after the push, only setting `should_run=true` when a genuinely new version header appears.

### Key findings

| Area | Assessment |
|------|-----------|
| **Correctness** | `comm -23` on sorted output correctly isolates new version headers. Both extraction paths (`get_changelog_versions` and inline `git show | sed`) use the same pattern `^## \([^ ]*\).*`. |
| **Edge cases** | Missing `HEAD~1` (orphan/first commit), zero SHA (`workflow_dispatch`), and git diff failures all fall through safely to conservative behavior (trigger rather than skip). Reordering entries produces no false positives. |
| **Security** | No injection vectors. All variables are CI-controlled. No eval or untrusted input. |
| **Workflow alignment** | `create-release.yml` passes `BEFORE`/`AFTER` correctly from `github.event.before` and `github.sha`. |

### Verdict

**APPROVED** — no blocking or non-blocking issues.

slopfairy (dismissed):
## Verdict: APPROVE

The change correctly gates release builds on genuinely new version headers rather than any changelog edit.

### ✅ What was checked

- **Correctness**: `comm -23` correctly isolates version headers present in the new file but absent in the `BEFORE` commit. Both `get_changelog_versions` and the inline `git show | sed` pipeline use the same `^## \([^ ]*\).*` extraction pattern, so comparisons are consistent.
- **Edge cases — handled**: `HEAD~1` missing (orphan branch, initial commit) → `git show` error suppressed by `2>/dev/null || true`, `versions_before` empty, first version triggers — safe conservative behavior. `BEFORE` set to zero SHA (workflow_dispatch) → same fallback applies. `BEFORE` properly set (push event) → correct range comparison.
- **Edge cases — no reordering false positives**: `comm -23` operates on sorted input, so reordering existing entries does not produce a false "new" version.
- **Security**: No injection vectors. Variables are CI-controlled. No eval or untrusted input execution.
- **Workflow alignment**: The gate step in `create-release.yml` passes `BEFORE: ${{ github.event.before }}` and `AFTER: ${{ github.sha }}`, which aligns with the script's expectations for both push and `workflow_dispatch` triggers.

### 🔴 Blocking

None.

### 🟡 Non-blocking

None.

slopfairy:
## Re-review complete — **APPROVED**

Reviewed the new push on PR #85 (commit `5c9f4a8`).

### What changed since last review

A second commit was pushed: `fix(ci): triage.yml uses inputs resolver for runs-on`. The original `tool/release.sh` gate change is unchanged.

### Review of `.github/workflows/triage.yml`

The `env` context is not available at `runs-on` evaluation time in GitHub Actions. This commit extracts the runner value into a dedicated `inputs` job, then all three jobs reference it via `needs.inputs.outputs.runner` — the same pattern used by `ci.yml` and `create-release.yml`. The simplified resolution (`${{ env.DEFAULT_RUNNER }}` without the `inputs.runner ||` fallback) is correct since triage.yml doesn't accept `workflow_dispatch` inputs.

### Verdict: **APPROVED**

No blocking or non-blocking issues. Both commits are correct and follow project conventions.

slopfairy (approved):
---
## Verdict: APPROVE
Second commit addresses the same `env`-in-`runs-on` issue found in triage.yml. Clean fix following established project patterns.

### 🔴 Blocking
None.

### 🟡 Non-blocking
None.

### ✅ Looks good
- `triage.yml`: `inputs` job resolver pattern matches `ci.yml` and `create-release.yml` exactly. The simpler resolution `${{ env.DEFAULT_RUNNER }}` (vs `${{ inputs.runner || env.DEFAULT_RUNNER }}`) is correct — triage.yml doesn't accept `workflow_dispatch` inputs.
- `tool/release.sh`: unchanged from first commit, already reviewed and approved.
- No security concerns (no new permissions, no untrusted input in evaluative context).

PR #93, opened 2026-06-10
fix: release gate tag-check fallback for workflow_dispatch

chaudharydeepanshu:
Gate BEFORE is empty on workflow_dispatch. Was comparing against HEAD~1 which already had the version. Now checks if the tag exists instead.

### DIFF
--- a/tool/release.sh
+++ b/tool/release.sh
@@ -461,21 +461,34 @@ cmd_gate() {
   fi
 
   if grep -Fqx "$target_file" <<< "$changed_files"; then
-    # Only trigger if a NEW version header was added, not just edits.
-    local versions_after versions_before new_version
+    local versions_after new_version
     versions_after=$(get_changelog_versions "$target_file")
-    versions_before=$(git show "${BEFORE:-HEAD~1}:$target_file" 2>/dev/null \
-      | sed -n 's/^## \([^ ]*\).*/\1/p' || true)
-    new_version=$(comm -23 <(echo "$versions_after" | sort) <(echo "$versions_before" | sort) | head -1)
+
+    if [[ -z "${BEFORE:-}" ]]; then
+      # No before-SHA (workflow_dispatch, force-push, new branch).
+      # Fall back to tag check: if the top version in the changelog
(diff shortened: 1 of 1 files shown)


slopfairy:
## Review: PR #93 — fix: release gate tag-check fallback for workflow_dispatch

**Verdict: APPROVE** ✅

### Analysis

**Problem**: On `workflow_dispatch`, `BEFORE` is empty. The old fallback `HEAD~1` could point to a commit that already contains the new version in the changelog, causing `comm -23` to find zero new versions and the gate to incorrectly skip the release.

**Fix**: When `BEFORE` is empty, check whether the top changelog version already has a corresponding `v{version}` git tag. If the tag exists, the version was already released and is skipped; if no tag exists, it is a new unreleased version and the gate fires.

**Cross-references verified**:
- Tag format `v{version}` confirmed at line 85 and line 537 of `release.sh`
- `get_changelog_versions | head -1` is an established pattern (line 530)

**Edge cases**:
- `BEFORE` unset + `git diff` fails → correctly enters tag-check branch
- `BEFORE` set + `git diff` fails → falls through to existing diff-based path (handled by `|| true`)
- Top version already tagged → `git rev-parse` succeeds → skips
- Top version untagged → `git rev-parse` fails → triggers release

No blocking issues. The shell quoting, error suppression, and branch separation are all correct. The comments clearly document the rationale.

slopfairy (approved):
## Verdict: APPROVE

The change correctly handles the `workflow_dispatch` edge case where `BEFORE` is empty by falling back to a tag-existence check instead of the broken `HEAD~1` comparison. The logic is consistent with the rest of the script's tag conventions and existing idioms.

### ✅ Looks good

- Tag format `v{version}` is verified (lines 85, 537)
- `get_changelog_versions | head -1` is an established pattern (line 530)
- Both `BEFORE`-empty and `BEFORE`-set paths are cleanly separated with no overlap
- All edge cases (empty changelog, existing tag, failed git commands) are handled correctly
- Comments clearly document intent

PR #110, opened 2026-06-21
ci: add changelog version check

chaudharydeepanshu:
Adds `tool/release.sh --check-versions` and wires it as a PR gate (both lanes, every PR) + a release-workflow backstop (releasing lane).

**Rule:** only the TOP heading of a changelog may be untagged (the release candidate). Every heading below must have its git tag, or a `<!-- release: no-tag -->` directive that still genuinely exists (GitHub Release or pub.dev). More than one new version = a heading faking a release it never got → blocked.

**Expected on this very PR:** the new `Changelog Versions` check goes **red** — `CHANGELOG.md` currently lists `## 2.0.2` below `## 2.1.0` with no `v2.0.2` tag (a leftover from the old mirror regen). That's the gate working. The fix (fold 2.0.2 into 2.1.0, retire the mirror regen) lands in this same PR before merge.

### DIFF
--- a/tool/release.sh
+++ b/tool/release.sh
@@ -76,6 +76,13 @@ Modes:
                             Env: BRANCH.
                             Outputs: tag, version, has_release.
 
+  --check-versions          Fail if the changelog adds >1 unreleased
+                            version. Only the top may be untagged; every
+                            heading below must have its git tag, or a
+                            '<!-- release: no-tag -->' directive AND still
+                            exist (GitHub Release or pub.dev).
+                            Env: BRANCH. Exit 1 on a faulty changelog.
+
   --github-notes TAG        Print GitHub Release notes to stdout.
   --update-tag-hashes TAG   Stamp asset hashes into the tag (post-upload).
   --stamp-changelog TAG     Build the filtered CHANGELOG.md for pub.dev.
@@ -123,7 +130,10 @@ extract_entry() {
   local file="$1" version="$2"
   local escaped
   escaped=$(printf '%s' "$version" | sed 's/[][\\.^$*+?(){}|]/\\&/g')
-  awk "/^## ${escaped}($| )/{found=1; next} /^## /{found=0} found" "$file"
+  # Drop reserved directive lines (<!-- release: ... -->) so they never
+  # reach the published changelog — they are internal tooling markers.
+  awk "/^## ${escaped}($| )/{found=1; next} /^## /{found=0} found" "$file" \
+    | sed '/<!--[[:space:]]*release:.*-->/d'
 }
 
 # List every "## version" heading in a changelog file, one per line.
@@ -252,24 +262,42 @@ JQ
 #
 # ── pub.dev changelog algorithm ───────────────────────────────────
 #
+# This filters the source changelog for the pub.dev tarball. It is
+# keyed PURELY on pub.dev publish status — NOT on git tags. (Whether a
+# version is real at all — has a git tag / GitHub release — is a
+# SEPARATE concern enforced before release by --check-versions; by the
+# time we reach here every heading is already a real, tagged version.)
+#
+# The case it handles: a version can be real and git-tagged yet never
+# pushed to pub.dev — you cut a GitHub release for issue-testing or a
+# `git: ref:` install and skipped/deferred pub.dev. It must NOT show as
+# a standalone pub.dev "## section" (pub.dev never had that version),
+# but its notes still matter — so they fold into the next published
+# version instead.
+#
 # Step 0 — Pick the source file.
 #   Stable (no "-") → CHANGELOG.md.  Pre (has "-") → CHANGELOG.pre.md.
 #
 # Step 1 — Classify each "## heading" as YES or NO.
-#   YES = it's already on pub.dev, OR it's the version being deployed.
-#   NO  = anything else (unpublished intermediate work).
+#   YES = already on pub.dev, OR the version being deployed right now
+#         (about to land on pub.dev this run).
+#   NO  = a real, tagged version that simply isn't on pub.dev.
 #
 #   YES → gets its own ## section in the output.
-#   NO  → folded into a collapsible under the nearest YES above it.
+#   NO  → folds into a collapsible under the nearest YES ABOVE it (the
+#         newer one). Why newer: the changelog is newest-first and
+#         releases are cumulative, so the next-newer published version
+#         already contains this NO version's changes — that's the
+#         pub.dev release its notes belong under.
 #
-# Pass 1 — Top to bottom, build the sections.
-#   YES versions get ## headings. Consecutive NO versions buffer into
-#   the previous YES version's "unpublished changes" collapsible.
+# Pass 1 — Top to bottom (newest→oldest), build the sections.
+#   YES versions get ## headings. Consecutive NO versions buffer, then
+#   attach to the previous YES (the newer one) when the next YES hits.
 #
 # Pass 2 — Bottom to top, add the commit lists.
 #   Each YES version gets a <details>Commits since vPREV</details>,
-#   where PREV is the YES version directly below it. The bottom-most
-#   YES version uses "Commits since initial".
+#   where PREV is the YES version directly below it (older). The
+#   bottom-most YES version uses "Commits since initial".
 #
 # ── Commit-range rule (same type only) ────────────────────────────
 #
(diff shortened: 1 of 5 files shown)


slopfairy:
posted my review below 👇

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — CHANGELOG.md `## 2.0.2` has no tag and will cause the new check to fail

1. `CHANGELOG.md:62` — The `## 2.0.2` heading sits below the top `## 2.1.0` release candidate with no `v2.0.2` git tag and no `<!-- release: no-tag -->` directive. The new `Changelog Versions` check this PR introduces will correctly flag it. As noted in the PR description, fold 2.0.2's entry into 2.1.0 (or remove it). Also, the mirror-regen instructions in the CHANGELOG.md header comment (lines 11-13) should be retired or at least noted as deprecated — regenerating CHANGELOG.md from CHANGELOG.pre.md without updating the corresponding tags will always trigger this check.

✅ **Checked**
- `tool/release.sh` `cmd_check_versions`: logic is sound — exempts the top heading, verifies tags on every heading below, handles no-tag directives with dual-source verification (GitHub Release + pub.dev), skips cleanly on missing files and single-version files
- `extract_entry` directive stripping (line 136): correctly removes `<!-- release: ... -->` lines from published changelog output without affecting the source file
- `.github/workflows/pr-lint.yml` `changelog-check` job: `permissions: contents: read` is appropriate for a PR gate; checks both lanes correctly; `fetch-tags: true` + `fetch-depth: 0` provides the tag data the script needs
- `.github/workflows/create-release.yml` backstop: added before the gate job — correct defense-in-depth placement; if a faulty merge skips the PR check, the release workflow still refuses to tag
- `set -e` safety: the `gh release view ... && on_gh=true` pattern on line 889 is safe — bash exempts the left side of `&&` from `set -e`, so a missing release won't cause an unexpected exit
- The pub.dev changelog algorithm documentation (lines 263-300) now clearly separates the two orthogonal concerns: "is it a real release?" (`--check-versions`) vs "is it on pub.dev?" (the fold in §3)
- `CHANGELOG.pre.md` is clean — the top `## 2.1.0-dev.0` is exempt, and every heading below carries the `-dev.N` suffix matching its lane's tag convention

slopfairy:
posted my review below 👇

tool/release.sh
```diff
@@ -747,13 +775,162 @@ dependencies:
...
+  local parsed
+  parsed=$(awk '
+    /^## / { if (v != "") print v "\t" n; v = $2; n = 0; next }
+    /<!--[[:space:]]*release:.*no-tag.*-->/ { n = 1 }
+    END { if (v != "") print v "\t" n }
+  ' "$file")
+
+  local total
+  total=$(grep -c . <<< "$parsed" || true)
+  if [ "$total" -le 1 ]; then
```
slopfairy, line 859: `grep -c . <<< "$parsed"` on an empty `parsed` (zero headings in the changelog) produces "0" with exit code 1, caught by `|| true`. The comparison `[ "$total" -le 1 ]` handles this correctly since `total` resolves to `"0"`. Edge-case: if the changelog has only HTML comments and no `##` headings, the awk produces no output, which is handled the same way. This is correct.

However, consider making `total` always numeric with `total=$(grep -c . <<< "$parsed" || echo 0)` to be explicit about the fallback rather than relying on grep outputting "0" before failing. Low priority.

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — `CHANGELOG.md` `## 2.0.2` has no `v2.0.2` tag or no-tag directive; the check is doing its job, now fix the data

## Review of the new changelog-version gate

The code is clean. The `cmd_check_versions` function is well-structured, the comments document the design rationale clearly, and the security posture of both workflows is correct. Below is what was checked and what remains.

---

1. `CHANGELOG.md:62` — `## 2.0.2` is below the top heading (`## 2.1.0`) with no `v2.0.2` tag and no `<!-- release: no-tag -->` directive. A release is cut from the top heading only, so `2.0.2` would silently collapse into `2.1.0` — a heading faking a release it never got. The PR description acknowledges this as expected, and says the fix (fold 2.0.2 into 2.1.0, retire the mirror regen) will land in this PR before merge. Until the changelog is clean, the new `Changelog Versions` gate (correctly) blocks this PR — let it be the proof.

---

### What was reviewed

**Workflow: `pr-lint.yml`** — the new `changelog-check` job
- `permissions: contents: read` — correct for a PR-level gate that only queries tags, releases, and pub.dev
- `fetch-tags: true` on checkout, plus a belt-and-suspenders `git fetch --tags --quiet origin` in the script — tags will be present
- Runs both lanes (`BRANCH=dev` → prerelease, `BRANCH=prod` → stable) — every PR gates both
- `GH_TOKEN` usage is all read-only — safe at `contents: read`

**Workflow: `create-release.yml`** — the "Version sanity" backstop
- Runs in the `gate` job, after checkout but before `--gate` decides to proceed
- Push-triggered (`on: push`), not `pull_request_target` — no fork risk
- Uses `${{ inputs.branch || github.ref_name }}` for the ref — always the pushed branch, never attacker-controlled

**`tool/release.sh` — `cmd_check_versions` function**
- The awk parser correctly associates a `<!-- release: no-tag -->` directive with the version above it (the section the directive appears in)
- pub.dev versions are fetched lazily and only when at least one below-top heading carries the no-tag directive — avoids an unnecessary network call on every PR
- The `no-tag` escape hatch requires proof — a GitHub Release OR pub.dev publication — rather than being a blank cheque; well-thought-out
- Error messages are actionable: they tell you exactly what's wrong and how to fix it
- The function is idempotent and safe to rerun

**`tool/release.sh` — updated pub.dev changelog comments**
- The expanded comments in section 3 clarify the separation of concerns: `--check-versions` asks "does this version exist at all?" while `build_pubdev_changelog` asks "is it on pub.dev?" — orthogonal checks that were previously easy to conflate

**`tool/release.sh` — `extract_entry` directive stripping**
- The added `sed '/<!--[[:space:]]*release:.*-->/d'` pipe ensures internal directives never reach pub.dev — correct and needed now that directives exist

---

### Not yet covered

- The CHANGELOG.md STANDARD comment block (lines 7-53) still describes the old "regenerate the stable file from CHANGELOG.pre.md" mirror regen workflow. The PR description says retiring the mirror regen is part of the fix. Once 2.0.2 is folded into 2.1.0 and the mirror regen is removed, update that comment block to match the new reality.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — changelog version check correctly implemented, 2.0.2 folded into 2.1.0, mirror regen retired

1. `tool/release.sh:854` — the `grep -c` fallback already flagged in the open thread is still worth making explicit with `|| echo 0`; low priority, not blocking

✅ **Checked**
- `cmd_check_versions` logic: top exempt, tag verification via `git rev-parse`, no-tag directive parsing, pub.dev cross-check for phantom versions — all correct
- `extract_entry` strips `<!-- release: ... -->` directives from the pub.dev changelog with a line-based sed delete — correct and safe
- `.github/workflows/pr-lint.yml`: changelog-check job with `contents: read`, both lanes checked, results accumulated via `rc` — correct
- `.github/workflows/create-release.yml`: version sanity backstop runs before the gate, `cmd_check_versions` fetches tags itself as belt-and-suspenders — correct placement
- CHANGELOG.md: `## 2.0.2` heading removed, content folded into `## 2.1.0`; `## 0.5.9` added with no-tag directive — correctly addresses the detected fault
- CHANGELOG.pre.md: instructions updated for independent lane changelogs, no content drift — correct
- Security: no `pull_request_target` misuse, minimal permissions (`contents: read` on the check job), no secrets in fork-triggerable context
- `git fetch --tags` inside `cmd_check_versions` (line 841) means even a shallow checkout without `fetch-tags` self-heals

slopfairy:
posted my review below 👇

tool/release.sh
chaudharydeepanshu: Good call on the readability, but `|| echo 0` would actually double-count: `grep -c` prints `0` *and* exits 1 on no match, so `grep -c . <<< "" || echo 0` yields `"0\n0"` and `[ "$total" -le 1 ]` then throws `integer expression expected`. Verified locally. Went with an explicit empty guard instead (741e921), which gives the same "always numeric, no reliance on grep's exit behavior" you were after.

slopfairy: > @chaudharydeepanshu: "Good call on the readability, but `|| echo 0` would actually double-count: `grep -c` prints `0` *and* exits 1 on no matc…"

You are right — I missed that `grep -c` outputs `0` to stdout before exiting 1, so `|| echo 0` would indeed append a second `0` rather than replacing. The explicit `[ -n "$parsed" ]` guard at line 856 achieves exactly the same goal (always numeric, no reliance on exit behavior) without the double-count pitfall. Clean fix.

slopfairy (approved):
> [!TIP]
> Approved — changelog check logic is correct, `## 2.0.2` is folded, workflows are secure, prior thread resolved

✅ **Checked**
- `tool/release.sh:826-923`: `cmd_check_versions` correctly exempts the top heading, requires tags or `no-tag` + pub.dev proof for everything below. The awk captures version+notag accurately including the END-block fallback for the last heading.
- `tool/release.sh:854-858`: prior thread resolved — `total=0` + `[ -n "$parsed" ]` guard avoids grep exit-1 on empty input, cleaner than the `|| echo 0` pattern.
- `tool/release.sh:129-137`: `extract_entry` drops `<!-- release: ... -->` directives so they never reach pub.dev.
- `CHANGELOG.md`: `## 2.0.2` folded into `## 2.1.0`, `## 0.5.9` added with `<!-- release: no-tag -->` directive.
- `CHANGELOG.pre.md`: standards block updated to the independent-lane model, version sequence intact.
- `.github/workflows/pr-lint.yml:91-110`: `changelog-check` job uses `contents: read`, `fetch-depth: 0`, `fetch-tags: true`. Both lanes checked in one run. No `pull_request_target`.
- `.github/workflows/create-release.yml:69-72`: version sanity backstop runs before gate, correct `BRANCH` derivation from `inputs.branch || github.ref_name`.
- Dispatch `main()` includes `--check-versions` in the case block.

PR #111, opened 2026-06-21
ci: flatten README <picture> banner for the pub.dev tarball

chaudharydeepanshu:
pub.dev strips `<picture>`/`<source>` when sanitizing the README, so the dark/light banner renders blank on the package page. This adds a `--stamp-readme` mode to `tool/release.sh` that flattens the banner to its inner `` in the pub.dev tarball at publish time, the same way `--stamp-changelog` rewrites the changelog. `README.md` in the repo stays `<picture>`, so GitHub keeps the dark/light version.

Tested: the flatten keeps the tracking comment intact, removes only the banner's `<picture>`/`<source>` tags, and is idempotent.

Also corrected the mode count and list in the release.sh header comment, which had drifted (it was missing `--check-versions`).

Refs: dart-lang/pub-dev#5923, dart-lang/pub-dev#6363, google/dart-neats#383

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -238,6 +238,8 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: bash tool/release.sh --stamp-changelog ${{ needs.discover.outputs.tag }}
+      - name: Flatten README banner for pub.dev
+        run: bash tool/release.sh --stamp-readme
       - name: Ephemeral commit (clean git for pub publish)
         run: |
           git config user.name "github-actions[bot]"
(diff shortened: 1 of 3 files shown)


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — clean fix with careful awk that handles the comment/standalone edge correctly

✅ **Checked**
- `cmd_stamp_readme` awk matches `<picture>` only as a standalone tag on its own line, never inside comment prose. The `<source` guard runs only while `inpic=1`, so comment mentions of `<source>` (if any were added later) would not be stripped. Idempotent via the `grep` guard.
- Workflow ordering is correct: `--stamp-changelog` → `--stamp-readme` → ephemeral commit, so both mutations land together.
- Mode count updated from 7 to 9; `--check-versions` and `--stamp-readme` added to the header list and pipeline flow consistently.
- No new secrets exposure in the workflow. The step runs in the existing publish job with the same environment protection.
- README.md HTML comment clearly documents why the `<picture>` stays in the repo and when to remove this mode.

PR #112, opened 2026-06-21
ci: auto-close resolved issues with an opt-out guard

chaudharydeepanshu:
Label-driven auto-close so resolved issues sanitize themselves without manual babysitting.

Flow:
- You add `resolved` when an issue is done from your side. The reporter immediately gets the opt-out notice (`auto-close-arm.yml`).
- Day 4: a reminder. Day 7: close as completed (`auto-close.yml`, `actions/stale`). `remove-stale-when-updated: false`, so a thanks or other chatter never stops the close.
- Only an explicit `keep open: <reason>` from a non-maintainer stops it: the guard removes `resolved` and `bot-closing-soon` (taking it out of the close pool) and tags @chaudharydeepanshu and @slopfairy with who asked and the reason (`auto-close-guard.yml`).
- After close, comments stay open and nothing auto-reopens.

Two labels: `resolved` (you add) and `bot-closing-soon` (managed by `actions/stale` as its stale marker; bot-only, do not edit).

Validated with actionlint. Comment bodies are read via env vars, never interpolated into the run script. Issues only, never PRs.

### DIFF
--- a/.github/workflows/auto-close-guard.yml
+++ b/.github/workflows/auto-close-guard.yml
@@ -37,5 +37,7 @@ jobs:
             | cut -c1-280)
           [ -z "$reason" ] && reason='(no reason given, see the comment)'
           gh issue edit "$NUM" --repo "$REPO" --remove-label resolved
+          # may also carry the stale marker if it opted out during the close window
+          gh issue edit "$NUM" --repo "$REPO" --remove-label bot-closing-soon 2>/dev/null || true
           gh issue comment "$NUM" --repo "$REPO" --body \
             "$(printf '@chaudharydeepanshu @slopfairy kept open by @%s.\nReason: %s' "$WHO" "$reason")"
--- a/.github/workflows/auto-close-guard.yml
+++ b/.github/workflows/auto-close-guard.yml
@@ -1,0 +1,41 @@
+# Stops the auto-close the instant a non-maintainer replies `keep open`.
+# Removes `resolved` (which takes the issue out of auto-close.yml's pool),
+# flags `needs-review`, and tags the maintainers with who asked and why.
+# Only acts on open issues, so it never reopens a closed one.
+name: Keep open on opt-out
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+
+jobs:
+  keep-open:
+    if: >
+      github.event.issue.state == 'open' &&
+      github.event.issue.pull_request == null &&
+      contains(github.event.issue.labels.*.name, 'resolved') &&
+      github.event.comment.author_association != 'OWNER' &&
+      github.event.comment.author_association != 'MEMBER' &&
+      github.event.comment.author_association != 'COLLABORATOR'
+    runs-on: ubuntu-latest
+    steps:
+      - env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BODY: ${{ github.event.comment.body }}
+          WHO: ${{ github.event.comment.user.login }}
+          NUM: ${{ github.event.issue.number }}
+          REPO: ${{ github.repository }}
+        run: |
+          # Comment body via env var, never interpolated into the script (injection-safe).
+          grep -qiE 'keep open' <<< "$BODY" || exit 0
+          reason=$(printf '%s' "$BODY" | tr '\n' ' ' \
+            | sed -E 's/.*[Kk]eep [Oo]pen//' \
+            | sed -E 's/^[[:space:]:,.-]+//' \
+            | sed -E 's/[[:space:]]+/ /g' \
+            | cut -c1-280)
+          [ -z "$reason" ] && reason='(no reason given, see the comment)'
+          gh issue edit "$NUM" --repo "$REPO" --remove-label resolved --add-label needs-review
+          gh issue comment "$NUM" --repo "$REPO" --body \
+            "$(printf '@chaudharydeepanshu @slopfairy kept open by @%s.\nReason: %s' "$WHO" "$reason")"
--- a/.github/workflows/auto-close-guard.yml
+++ b/.github/workflows/auto-close-guard.yml
@@ -1,6 +1,6 @@
 # Stops the auto-close the instant a non-maintainer replies `keep open`.
-# Removes `resolved` (which takes the issue out of auto-close.yml's pool),
-# flags `needs-review`, and tags the maintainers with who asked and why.
+# Removes `resolved` (which takes the issue out of auto-close.yml's pool)
+# and tags the maintainers with who asked and why.
 # Only acts on open issues, so it never reopens a closed one.
 name: Keep open on opt-out
 on:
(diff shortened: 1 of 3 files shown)


slopfairy:
posted my review below 👇

.github/workflows/auto-close-guard.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          # Comment body via env var, never interpolated into the script (injection-safe).
+          grep -qiE 'keep open' <<< "$BODY" || exit 0
+          reason=$(printf '%s' "$BODY" | tr '\n' ' ' \
+            | sed -E 's/.*[Kk]eep [Oo]pen//' \
+            | sed -E 's/^[[:space:]:,.-]+//' \
+            | sed -E 's/[[:space:]]+/ /g' \
+            | cut -c1-280)
+          [ -z "$reason" ] && reason='(no reason given, see the comment)'
+          gh issue edit "$NUM" --repo "$REPO" --remove-label resolved --add-label needs-review
```
slopfairy: Suggest also removing `pending-auto-close` here. When the guard fires after the stale action has already labeled the issue as `pending-auto-close` (day 4-7 window), the label stays as a visual artifact even though the issue is now out of the auto-close pool.

```suggestion
          gh issue edit "$NUM" --repo "$REPO" --remove-label resolved --remove-label pending-auto-close --add-label needs-review
```

slopfairy (dismissed):
> [!TIP]
> Approved — well-designed workflows with proper security boundaries

1. `.github/workflows/auto-close-guard.yml:39` — also remove `pending-auto-close` when the guard fires, so the stale marker doesn't linger on an issue that's now out of the close pool

✅ **Checked**
- No `pull_request_target` triggers — all workflows use `issues`, `issue_comment`, and `schedule` events, safe from fork-based attacks
- Permissions scoped to minimum: `issues: write` on guard/arm, plus `pull-requests: write` only on auto-close (needed by `actions/stale`)
- `github.event.comment.body` read via env var (`$BODY`), never interpolated into expression syntax — injection-safe as described
- `actions/stale@v9` is pinned to a major version, a trusted GitHub action
- Guard condition correctly filters to open issues, non-PR comments, `resolved`-labeled, and non-maintainer authors
- `exempt-issue-labels: keep-open` provides a manual escape hatch for issues that should never auto-close
- `remove-stale-when-updated: false` ensures chatter cannot reset the close clock
- `days-before-pr-stale: -1` / `days-before-pr-close: -1` correctly disables the stale action for pull requests
- Reason extraction pipeline handles multi-line comments, trims whitespace, and caps at 280 characters

slopfairy:
posted my review below 👇

.github/workflows/auto-close-guard.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          # Comment body via env var, never interpolated into the script (injection-safe).
+          grep -qiE 'keep open' <<< "$BODY" || exit 0
+          reason=$(printf '%s' "$BODY" | tr '\n' ' ' \
+            | sed -E 's/.*[Kk]eep [Oo]pen//' \
+            | sed -E 's/^[[:space:]:,.-]+//' \
+            | sed -E 's/[[:space:]]+/ /g' \
+            | cut -c1-280)
+          [ -z "$reason" ] && reason='(no reason given, see the comment)'
+          gh issue edit "$NUM" --repo "$REPO" --remove-label resolved
```
slopfairy, line 39: Two problems with this line:

1. The guard should also add `needs-review` — the PR body describes the guard as flagging `needs-review` so maintainers can find issues where a user objected. Without it, the only signal is a comment mention, which is easily missed.

2. After day 4, the stale action labels the issue `bot-closing-soon`. When the user opts out, `resolved` goes away (stopping the close), but `bot-closing-soon` remains as a stale visual artifact. Remove it too.

```suggestion
          gh issue edit "$NUM" --repo "$REPO" --remove-label resolved --remove-label bot-closing-soon --add-label needs-review
```

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the guard never adds `needs-review` so maintainers have no systematic way to find opt-out issues

1. `.github/workflows/auto-close-guard.yml:39` — The guard removes `resolved` but does not add `needs-review`. The PR description states the guard should flag `needs-review` so objected issues are discoverable. Without it, the only signal is a comment @-mention, buried in the notification stream. The fix is on the inline suggestion above.
2. `.github/workflows/auto-close-guard.yml:39` — The guard also does not remove `bot-closing-soon`. After the stale action runs (day 4), the issue carries both `resolved` and `bot-closing-soon`. When a user opts out, only `resolved` is stripped; `bot-closing-soon` remains as a misleading artifact. The inline suggestion covers both fixes.

1. `.github/workflows/auto-close-arm.yml:11-22` — The arm workflow fires on every `labeled` event where the label equals `resolved`. If the label is removed and re-added (or if `resolved` is among other labels being added), a duplicate notice comment is posted. Consider a guard against re-posting.
2. `.github/workflows/auto-close.yml:15` — `pull-requests: write` is declared but PR handling is fully disabled (`days-before-pr-stale: -1`, `days-before-pr-close: -1`). The permission is unnecessary.

✅ **Checked**
- No `pull_request_target` — all events are `issues`, `issue_comment`, and `schedule`; fork-safe
- No custom secrets — only `secrets.GITHUB_TOKEN`
- Comment body read via env var, never interpolated into the run script — injection-safe
- `remove-stale-when-updated: false` — correct; chatter should not reset the close clock
- `only-labels: 'resolved'` with `days-before-pr-*: -1` — only issues are targeted, PRs excluded
- Author-association guard (`OWNER`, `MEMBER`, `COLLABORATOR`) correctly excludes maintainers from triggering the guard
- `actions/stale@v9` is a well-known, actively maintained action

The PR body mentions four labels (`resolved`, `keep-open`, `pending-auto-close`, `needs-review`) but the code uses `bot-closing-soon` instead of `pending-auto-close`, and `keep-open` is never referenced in any workflow. The code is the source of truth — the description just needs updating before merge.

chaudharydeepanshu:
Thanks @slopfairy. Applied the real fix: the guard now also removes `bot-closing-soon` on opt-out, so it can't linger as an artifact. Tightened `auto-close.yml` to `issues: write` only.

On `needs-review`: that label was intentionally dropped (redundant with the opt-out comment plus the @mention), so the guard doesn't add it by design. Updated the PR description to match the code, which now uses `resolved` + `bot-closing-soon` only.

Left the arm re-post edge as-is for the pilot; re-labeling is rare and it's a duplicate comment at worst.

slopfairy:
posted my review below 👇

slopfairy:
> @chaudharydeepanshu: "Applied the real fix: the guard now also removes `bot-closing-soon` on opt-out"

The `bot-closing-soon` removal closes the artifact gap — no zombie close timer after opt-out. `issues: write` on `auto-close.yml` is the right scope. Dropping `needs-review` keeps the label set smaller and avoids a stale/close race. Sounds good.

On the re-post edge: worth a note in the PR thread if it ever surfaces, but not worth blocking over.

slopfairy (approved):
> [!TIP]
> Approved — both previous threads resolved; `bot-closing-soon` guard fix applied and permissions tightened

✅ **Checked**
- `auto-close-guard.yml:41` — now removes `bot-closing-soon` on opt-out with `|| true` suppression for the no-label case
- `auto-close.yml:14` — permissions tightened to `issues: write` only
- No pull_request_target, no secrets in fork-triggerable context — all three workflows trigger on `issues` / `issue_comment` (repo context only)
- Comment bodies read exclusively via env vars, never interpolated into run scripts
- `pull_request == null` guard in auto-close-guard.yml + `days-before-pr-*: -1` in auto-close.yml correctly limit scope to issues
- `actions/stale@v9` parameters (`only-labels`, `stale-issue-label`, `close-issue-reason`) are valid for the v9 API

PR #117, opened 2026-06-23
ci: declare repo labels as code with a sync workflow

chaudharydeepanshu:
Declares the repo's 12 labels in `.github/labels.json` (single source of truth) and adds `.github/workflows/labels.yml` to sync GitHub to it on push to `dev` (create / update / delete strays). Stops a label a workflow references from silently not existing — the `dependencies` miss that just broke Dependabot.

Follows the existing workflow standards: `permissions: {}` + least-privilege per job, `DEFAULT_RUNNER` + `inputs` resolver, repo-guard `if`, pinned `actions/checkout` with `persist-credentials: false`, inline `gh api` (no third-party action, same posture as `triage.yml`), bash kept 3.2-clean.

Verified locally: JSON/YAML valid, `lint_shell.sh` + actionlint clean, and a dry-run against the live labels shows **0 changes** — so the first run after merge is a guaranteed no-op.

### DIFF
--- a/.github/labels.json
+++ b/.github/labels.json
@@ -1,0 +1,14 @@
+[
+  { "name": "bug",              "color": "D93F0B", "description": "Something broken" },
+  { "name": "feature",          "color": "0E8A16", "description": "New capability" },
+  { "name": "ci",               "color": "BFD4F2", "description": "Build, CI, workflows" },
+  { "name": "web",              "color": "1D76DB", "description": "Web or WASM specific" },
+  { "name": "upstream",         "color": "5319E7", "description": "Needs upstream pdf_oxide change" },
+  { "name": "dependencies",     "color": "0366D6", "description": "Dependency version bumps" },
+  { "name": "ready-to-test",    "color": "FBCA04", "description": "Triggers full test suite" },
+  { "name": "release",          "color": "6F42C1", "description": "Release promotion PR" },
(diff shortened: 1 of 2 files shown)


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — follows the repo's workflow conventions exactly and adds label management as code cleanly

✅ **Checked**
- `labels.json` is valid JSON, all 12 labels consistent with what the repo already has (0-change dry-run verified)
- Trigger is `push` to `dev` — no `pull_request_target`, no fork exposure, no untrusted event data in `run:`
- `permissions: {}` at top, least-privilege per-job (`contents: read` + `issues: write` only on `sync`)
- `actions/checkout` SHA-pinned with `persist-credentials: false`, no third-party label actions, inline `gh api` only
- Repo-guard `if` on both jobs prevents forks from syncing their own labels
- `inputs` pattern, `DEFAULT_RUNNER`, concurrency, `timeout-minutes` all match triage.yml and other workflows
- `secrets.GITHUB_TOKEN` flows through `env:` on the step, never interpolated into the `run:` block
- Bash is 3.2-clean (process substitution, `set -euo pipefail`, POSIX loop constructs); passed actionlint + shell lint
- Authoritative delete loop will clean up strays — loud-failure on API errors (no `|| true`), which is appropriate for a trusted push trigger

PR #113, opened 2026-06-22
ci: harden CI to zizmor-auditor zero with a supply-chain SSOT

chaudharydeepanshu:
## What

Drives the entire `.github` tree to `zizmor --persona=auditor` exit-0, and replaces the home-rolled workflow scanner with a real zizmor + actionlint gate in `pr-lint.yml` so the bar stays enforced on every PR.

## Highlights

- **Supply-chain SSOT.** Every pinned version + sha256 lives in `tool/versions.env`; third-party binaries download fail-closed through `tool/fetch_verified.sh` (missing/mismatched hash = refusal).
- **Two privileged workflows, hardened + documented.** triage (`pull_request_target`) and retry (`workflow_run`) are the only way GitHub grants fork-PR write on a public repo; each is hardened to the doctrine in `docs/ARCHITECTURE.md` (Workflow security) — no PR checkout, fork data env-only, least-privilege per-job tokens, repo-guarded, SHA-pinned — and carries exactly one documented `# zizmor: ignore[dangerous-triggers]`.
- **Upgrade bot.** `upgrade-check.yml` now bumps everything Dependabot can't see (SDK, binaries, zizmor, actionlint) AND refreshes the committed lockfiles, in one reviewed PR.
- **Determinism.** `pubspec.lock` committed (kept fresh by the bot); test pass/fail reconciled from the JSON reporter, not console strings.
- Action SHA-pinning throughout; `dtolnay/rust-toolchain` → runner rustup; `softprops/action-gh-release` → `gh release upload`.
- CODEOWNERS gates the build / release / supply-chain surface.

## Testing

Add `ready-to-test` for the full cross-target matrix. On the first external fork PR, confirm triage actually labels (the PR-write permission is research-correct but unproven on a live fork until then).

PR #120, opened 2026-06-23
fix(ci): re-supply git push auth in release.sh

chaudharydeepanshu:
## What

`create-release.yml`'s "Discover versions" job was failing in ~20s with:

```
fatal: could not read Username for 'https://github.com': No such device or address
```

## Why

The CI hardening (#113) set `persist-credentials: false` on the release checkouts. That strips the token `git push` needs. `gh release create` survived (it reads `GH_TOKEN`), but `release.sh`'s staging-branch push and tag force-push had nothing to authenticate with. v2.1.0 was the last release cut before #113 landed; every release since has been broken at this step.

## Fix

`git_ci_identity()` (the shared setup both push paths call) now runs `gh auth setup-git` when `GH_TOKEN` is set, wiring the token in as a git credential helper just before the pushes.

Considered flipping the discover job to `persist-credentials: true` instead. Ran the repo's auditor-persona invocation and it produces a medium `artipacked` finding, which breaks zizmor-zero and would need an avoidable ignore. Keeping `persist-credentials: false` and re-supplying via `gh` keeps the workflow hardened and zizmor clean.

Fixes both `--discover` and `--update-tag-hashes` push paths. Local: `bash -n` + `tool/lint_shell.sh` pass; zizmor unchanged at zero.

### DIFF
--- a/tool/ci/release.sh
+++ b/tool/ci/release.sh
@@ -125,10 +125,18 @@ gh_output() {
   echo "  output: $1=$2"
 }
 
-# Configure the git identity used for CI commits.
+# Configure the git identity + push auth used for CI commits.
 git_ci_identity() {
   git config user.name  "github-actions[bot]"
   git config user.email "github-actions[bot]@users.noreply.github.com"
+
+  # The release checkout uses persist-credentials:false (zizmor hardening leaves
+  # no token in git config for later steps to leak), so these pushes have no auth.
+  # Wire gh's token in as a credential helper — the same GH_TOKEN `gh release`
+  # already uses. Skipped on a local dry run, where the push uses your own creds.
+  if [ -n "${GH_TOKEN:-}" ]; then
+    gh auth setup-git
+  fi
 }
 
 # Extract one version's entry (body only, heading excluded) from a


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — minimal, targeted fix that re-supplies push auth only where needed, preserving the `persist-credentials: false` hardening from #113

✅ **Checked**
- `git_ci_identity()` now runs `gh auth setup-git` when `GH_TOKEN` is set, covering both `--discover` (staging-branch push + tag force-push) and `--update-tag-hashes` (tag force-push)
- The guard `[ -n "${GH_TOKEN:-}" ]` correctly skips the credential setup on local dry runs, falling back to the user's own credentials
- `persist-credentials: false` remains on all release checkouts — the workflow hardening from #113 is intact and zizmor stays at zero
- `gh auth setup-git` is idempotent and safe to call repeatedly; the credential helper lives only for the job's lifetime, no cross-job leakage
- The `publish` job's manual `git config` (lines 286-287 of the workflow) does not need the fix — it only creates a local ephemeral commit with no push
- `GH_TOKEN` is already passed to both affected job steps (`discover` at line 122, `upload-assets` at line 236)

PR #121, opened 2026-06-24
ci: unblock the release pipeline + supply-chain & workflow hardening

chaudharydeepanshu:
Carries the release-critical fix plus a hardening pass on the CI and supply-chain tooling, all on one branch per request.

## Release-critical (ships v2.1.1)

Two regressions from the #113 hardening left the release pipeline broken at its first two steps. #120 fixed the push auth; this fixes the next: `gh release upload` runs before its checkout, so it had no repo to infer, and `GH_REPO` names it. Together they unblock the pipeline end to end.

## Supply-chain hardening

Suggestions on the pin and upgrade tooling. Every item was checked against the code before acting.

### Applied
- **Injection on write.** `set_kv` shape-checks each value (version or 64-hex sha) before writing the sourced `versions.env`, and the Flutter version is gated to exact semver before it reaches `sed` (a separate sink, where GNU sed's `e` command is RCE). A shell-lint check enforces that only `set_kv` writes the file.
- **Tamper + availability.** `verify-pinned` HEADs then re-hashes each pin (404/410 = pruned, 200 with a changed hash = repoint, transient = warn). A HEAD-only `check-availability` runs from pr-lint on every PR, so a stale pin fails the PR even when the daily cron is disabled after 60 idle days.
- **fetch_verified.** Fail-closed on a 64-hex allowlist (was a placeholder blocklist); bounds the download on stall rather than a flat timeout.
- **Fail loud.** A blocked pin bump exits nonzero instead of passing silently.
- **Portable cargo config.** awk in place of a GNU-only `sed` insert that broke on macOS.
- **Verified publish.** The publish job uses the hash-pinned fvm capability, not an unverified `pub global activate`.
- **Upgrade radar split.** Security-sensitive pin bumps get their own PR (`chore/pins`), reviewed undiluted by lockfile churn; lockfiles refresh on `chore/lockfiles`.

### Declined, with the reason
- **yq command-word + sh allowlist:** no current false-positive, and allowing `sh` weakens the Windows-safety this check exists for.
- **dangling-path lint:** false-positives on `.dart_tool/` substrings and example paths in docs; a precise boundary needs the non-portable regex the file bans.
- **mirror pins to our own release:** hundreds of MB per bump plus a new failure mode. The cheap half (detect pruning) is applied instead, via the PR-time availability check above.
- **doc/code tool-ban grep:** can't tell a scoped-true "the hook uses no jq" from a global-false claim.

## Workflow hardening

A full inspection found 19 jobs across five workflows (ci, create-release, full-test, pr-lint, upgrade-check) with no `timeout-minutes`, falling back to the 6-hour default. Every job is now bounded: 5 for trivial orchestration, 10 to 20 for checkout and network jobs, 30 to 45 for the heavy download and compile jobs. Checkout hardening (SHA-pinned, persist-credentials:false) and per-job permissions were already uniform.

## Dev-tooling gates

Three local gates claimed to enforce something they didn't:

- **The gitignore pre-commit hook never fired.** `git check-ignore` consults the index by default, so a force-added file (the only way to stage an ignored one) reads as tracked, not ignored, and the guard saw nothing. `--no-index` applies the rules regardless of index state; path handling is now NUL-safe end to end.
- **`make build-native` passed on a failed compile.** The recipe swallowed the build hook's exit with `|| true`; since a no-match test and a compile failure both exit nonzero, a broken native build looked green. It now tees the output and fails on the dart-native tooling's build-failure markers.
- **The conventional-commit type list was duplicated** in the commit-msg hook and pr-lint, free to drift apart. Both now read `tool/commit-types.txt`.

- **The `dart:io` test guard didn't say why anything was exempt.** `test-guards` listed its exemptions as positional `grep -v` paths in the Makefile, so the reason a test was allowed `dart:io` lived nowhere and a rename would silently drop a file from the list. Each exempt test now carries an `io-exempt: <reason>` comment and the guard greps for the marker. The byte-grep guard stays positional on purpose: `sublist(0, 5)` and `isNot(contains ...)` are always-fine idioms (not per-file opt-outs), and `pdf_builder_battery` has 15 such lines where one file exclusion beats marking each.

## Review gating

CODEOWNERS gated `.github/`, `/tool/`, and lockfiles to the owner but let `/vendor/` fall to the maintainers team via `*`. A submodule pointer bump pulls in that commit's `build.rs`, which cargo runs at compile time in CI, so the gitlink bump is the merge gate that matters; it's now owner-gated. The per-file patterns an audit usually adds (`/vendor/**/build.rs`, `Cargo.lock`) are intentionally omitted: those files live inside the submodule repos, where a parent CODEOWNERS can't match them, so they're gated by the submodule repos' own CODEOWNERS.

## Docs

Brought up to date with the above: ARCHITECTURE principle 3 (bash uses jq, the hook is Dart), SECURITY's self-computed-pin note (now references verify-pinned/check-availability), UPDATING's upgrade flow (the pins/lockfiles split), and upgrade.sh's usage header (the new modes). Labels gained a consistent description schema and an (auto)/(manual) applier tag.

Local gates green: shellcheck, lint_shell (bash 3.2 + BSD/GNU), zizmor `--persona=auditor` at zero with no new ignores, actionlint.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -21,12 +21,15 @@ if printf '%s\n' "$msg" | grep -qE '^Merge '; then
   exit 1
 fi
 
-if ! printf '%s\n' "$msg" | grep -qE '^(feat|fix|docs|chore|ci|test|refactor|perf|build|deps)(\(.+\))?!?: .+'; then
+# Single source of truth for the type list: tool/commit-types.txt (pr-lint.yml
+# validates PR titles against the same file). cwd is the repo root during hooks.
+types=$(paste -sd'|' tool/commit-types.txt)
+if ! printf '%s\n' "$msg" | grep -qE "^(${types})(\(.+\))?!?: .+"; then
   echo "ERROR: Commit message must follow Conventional Commits format."
   echo ""
   echo "  <type>[optional scope]: <description>"
   echo ""
-  echo "  Types: feat fix docs chore ci test refactor perf build deps"
+  printf '  Types: %s\n' "$(paste -sd' ' tool/commit-types.txt)"
   echo "  Examples:"
   echo "    feat: add watermark support"
   echo "    fix(web): handle empty PDF edge case"
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -2,9 +2,13 @@
 # Reject commits that add files matching .gitignore patterns.
 # Catches accidental `git add -f` of ignored files.
 
-bad_files=$(git diff --cached --name-only --diff-filter=A | while read f; do
-  git check-ignore --quiet "$f" 2>/dev/null && echo "$f"
-done)
+# --no-index is load-bearing: a force-added file is already in the index, and
+# without it git check-ignore treats an indexed path as tracked and reports
+# nothing, so the hook would silently never fire. NUL end to end
+# (core.quotepath=false stops octal-escaping, -z stops newline-splitting) keeps
+# paths with spaces or newlines intact; --stdin avoids a non-POSIX read -d '' loop.
+bad_files=$(git -c core.quotepath=false diff --cached --name-only --diff-filter=A -z \
+  | git check-ignore --no-index --stdin -z 2>/dev/null | tr '\0' '\n')
 
 if [ -n "$bad_files" ]; then
   echo "ERROR: These newly-added files are gitignored:"
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -26,3 +26,12 @@
 # (fetch_verified.sh), the publish pipeline + the hash-writing bump bot
 # (tool/ci/). A non-owner edit here is a supply-chain or release risk.
 /tool/                     @chaudharydeepanshu
+
+# Vendored engine submodules: bumping a gitlink pulls in that commit's build.rs
+# and build-dependencies, which cargo runs at compile time (before any test) in
+# CI. Without this, a bump falls to the maintainers team via `*`; gate the
+# pointer bump to the owner instead. Per-file rules like /vendor/**/build.rs are
+# deliberately absent: those files live inside the submodule repos, not this one,
+# so CODEOWNERS here can't match them. The submodule repos' own CODEOWNERS gate
+# their build scripts and lockfiles.
+/vendor/                   @chaudharydeepanshu
--- a/.github/actions/capabilities/rust/action.yml
+++ b/.github/actions/capabilities/rust/action.yml
@@ -96,11 +96,20 @@ runs:
         fi
         mkdir -p "$HOME/.cargo"
         CFG="$HOME/.cargo/config.toml"
-        # A fresh runner has no config.toml, so the append creates it. Guard
-        # the rare pre-existing case: a second [build] header is a TOML error,
-        # so insert the wrapper UNDER the existing table instead.
+        # A fresh runner has no config.toml, so the append creates it. Guard the
+        # rare pre-existing case (a second [build] header is a TOML error) by
+        # inserting the wrapper UNDER the existing table. Done with awk, not a
+        # GNU sed insert: the one-line append form is GNU-only (BSD wants a
+        # two-line backslash form), so it broke on macOS runners. SCCACHE_BIN
+        # rides ENVIRON, not awk -v, so a path is never backslash-interpreted.
+        # (An env var via GITHUB_ENV would be simpler but zizmor flags every
+        # dynamic GITHUB_ENV write as an injection vector — a file write is clean.)
         if [ -f "$CFG" ] && grep -q '^\[build\]' "$CFG"; then
-          sed -i.bak '/^\[build\]/a rustc-wrapper = "'"$SCCACHE_BIN"'"' "$CFG" && rm -f "$CFG.bak"
+          sk_w="$SCCACHE_BIN" awk '
+            BEGIN { w = ENVIRON["sk_w"] }
+            { print }
+            /^\[build\]/ && !seen { print "rustc-wrapper = \"" w "\""; seen=1 }
+          ' "$CFG" > "$CFG.tmp" && mv "$CFG.tmp" "$CFG"
         else
           printf '\n[build]\nrustc-wrapper = "%s"\n' "$SCCACHE_BIN" >> "$CFG"
         fi
--- a/.github/actions/capabilities/rust/action.yml
+++ b/.github/actions/capabilities/rust/action.yml
@@ -56,6 +56,10 @@ runs:
           # connection reset). cargo's default is 3 retries; 10 rides out a
           # short registry blip instead of failing the whole build.
           echo "CARGO_NET_RETRY=10"
+          # HTTP/2 multiplexing intermittently fails crates.io with a framing
+          # layer error (curl 16) that the retry above doesn't catch; HTTP/1.1
+          # avoids that class.
+          echo "CARGO_HTTP_MULTIPLEXING=false"
         } >> "$GITHUB_ENV"
         # Submodule pins arrive via env (hoisted from step outputs). Heredoc
         # with a random delimiter so a stray newline can't inject entries.
--- a/.github/actions/capabilities/rust/action.yml
+++ b/.github/actions/capabilities/rust/action.yml
@@ -96,14 +96,9 @@ runs:
         fi
         mkdir -p "$HOME/.cargo"
         CFG="$HOME/.cargo/config.toml"
-        # A fresh runner has no config.toml, so the append creates it. Guard the
-        # rare pre-existing case (a second [build] header is a TOML error) by
-        # inserting the wrapper UNDER the existing table. Done with awk, not a
-        # GNU sed insert: the one-line append form is GNU-only (BSD wants a
-        # two-line backslash form), so it broke on macOS runners. SCCACHE_BIN
-        # rides ENVIRON, not awk -v, so a path is never backslash-interpreted.
-        # (An env var via GITHUB_ENV would be simpler but zizmor flags every
-        # dynamic GITHUB_ENV write as an injection vector — a file write is clean.)
+        # Insert under [build] (a second header is a TOML error) with awk, not a
+        # GNU one-line sed append; that form broke on macOS. Pass the path via
+        # ENVIRON, not awk -v, which would interpret a backslash in it.
         if [ -f "$CFG" ] && grep -q '^\[build\]' "$CFG"; then
           sk_w="$SCCACHE_BIN" awk '
             BEGIN { w = ENVIRON["sk_w"] }
--- a/.github/actions/capabilities/rust/action.yml
+++ b/.github/actions/capabilities/rust/action.yml
@@ -104,7 +104,7 @@ runs:
             BEGIN { w = ENVIRON["sk_w"] }
             { print }
             /^\[build\]/ && !seen { print "rustc-wrapper = \"" w "\""; seen=1 }
-          ' "$CFG" > "$CFG.tmp" && mv "$CFG.tmp" "$CFG"
+          ' "$CFG" > "$CFG.tmp" && mv "$CFG.tmp" "$CFG" || rm -f "$CFG.tmp"
         else
           printf '\n[build]\nrustc-wrapper = "%s"\n' "$SCCACHE_BIN" >> "$CFG"
         fi
--- a/.github/labels.json
+++ b/.github/labels.json
@@ -1,14 +1,16 @@
 [
-  { "name": "bug",              "color": "D93F0B", "description": "Something broken" },
-  { "name": "feature",          "color": "0E8A16", "description": "New capability" },
-  { "name": "ci",               "color": "BFD4F2", "description": "Build, CI, workflows" },
-  { "name": "web",              "color": "1D76DB", "description": "Web or WASM specific" },
-  { "name": "upstream",         "color": "5319E7", "description": "Needs upstream pdf_oxide change" },
-  { "name": "dependencies",     "color": "0366D6", "description": "Dependency version bumps" },
-  { "name": "ready-to-test",    "color": "FBCA04", "description": "Triggers full test suite" },
-  { "name": "release",          "color": "6F42C1", "description": "Release promotion PR" },
-  { "name": "needs-info",       "color": "D876E3", "description": "Waiting on reporter" },
-  { "name": "wont-fix",         "color": "EEEEEE", "description": "Intentional or out of scope" },
-  { "name": "resolved",         "color": "1A7F37", "description": "Done from maintainer side; auto-closes after a grace period" },
-  { "name": "bot-closing-soon", "color": "CFD3D7", "description": "Auto-close in progress; bot-managed, do not edit" }
+  { "name": "bug",              "color": "D93F0B", "description": "Broken behavior (auto)" },
+  { "name": "feature",          "color": "0E8A16", "description": "New capability (auto)" },
+  { "name": "ci",               "color": "BFD4F2", "description": "Build, CI, or workflow change (auto)" },
+  { "name": "web",              "color": "1D76DB", "description": "Web or WASM specific (auto)" },
+  { "name": "upstream",         "color": "5319E7", "description": "Needs a pdf_oxide change (auto)" },
+  { "name": "dependencies",     "color": "0366D6", "description": "Dependency version bump (auto)" },
+  { "name": "upgrade-pins",     "color": "1D76DB", "description": "Pinned tool or binary bump, review the hashes (auto)" },
+  { "name": "upgrade-locks",    "color": "1D76DB", "description": "Lockfile refresh (auto)" },
+  { "name": "ready-to-test",    "color": "FBCA04", "description": "Triggers the full cross-target test run (manual)" },
+  { "name": "release",          "color": "6F42C1", "description": "Release promotion PR (manual)" },
+  { "name": "needs-info",       "color": "D876E3", "description": "Awaiting the reporter (manual)" },
+  { "name": "wont-fix",         "color": "EEEEEE", "description": "Intentional or out of scope (manual)" },
+  { "name": "resolved",         "color": "1A7F37", "description": "Fixed, auto-closes after a grace period (manual)" },
+  { "name": "bot-closing-soon", "color": "CFD3D7", "description": "Auto-close pending, do not edit (auto)" }
 ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
   inputs:
     name: Resolve inputs
     runs-on: ubuntu-24.04
+    timeout-minutes: 5
     permissions: {}
     outputs:
       runner: ${{ steps.resolve.outputs.runner }}
@@ -42,6 +43,7 @@ jobs:
     name: Detect changes
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 5
     permissions:
       contents: read  # checkout + paths-filter, read-only
     outputs:
@@ -127,6 +129,7 @@ jobs:
     if: always()
     needs: [inputs, changes, analyze, test, rust]
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 5
     permissions: {}
     steps:
       - name: Evaluate
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -54,6 +54,7 @@ jobs:
   inputs:
     name: Resolve inputs
     runs-on: ubuntu-24.04
+    timeout-minutes: 5
     permissions: {}
     outputs:
       runner: ${{ steps.resolve.outputs.runner }}
@@ -69,6 +70,7 @@ jobs:
     name: Release gate
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 5
     permissions:
       contents: read  # checkout + changelog gate, read-only
     outputs:
@@ -104,6 +106,7 @@ jobs:
       group: release-${{ needs.gate.outputs.version || github.run_id }}
       cancel-in-progress: true
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 20
     permissions:
       contents: write  # release.sh --discover creates the GitHub release
     outputs:
@@ -140,6 +143,7 @@ jobs:
           - { name: windows, runner: windows-2025-vs2026 }
     name: compile-${{ matrix.name }}
     runs-on: ${{ matrix.runner }}
+    timeout-minutes: 45
     permissions:
       contents: read  # checkout + build; artifacts use run storage, no API
     steps:
@@ -162,6 +166,7 @@ jobs:
     needs: [inputs, discover, compile]
     if: needs.discover.outputs.has_release == 'true'
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 15
     permissions:
       contents: write  # uploads binaries + stamps the release tag
     steps:
@@ -258,6 +263,7 @@ jobs:
     needs: [inputs, discover, upload-assets]
     if: needs.discover.outputs.has_release == 'true'
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 20
     environment: ${{ inputs.branch || github.ref_name }}
     permissions:
       contents: write  # ephemeral stamp commit + release-notes install snippet
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -218,6 +218,9 @@ jobs:
       - name: Upload binaries to the release
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # This step runs before any checkout, so gh has no .git to infer the
+          # target repo from. GH_REPO names it, or gh exits "not a git repo".
+          GH_REPO: ${{ github.repository }}
           TAG: ${{ needs.discover.outputs.tag }}
         # gh ships on the runner; the release already exists (discover
         # created it), so just upload the binaries, clobbering on rerun.
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -267,10 +267,8 @@ jobs:
           ref: '${{ needs.discover.outputs.tag }}'
           fetch-depth: 0
           persist-credentials: false
-      # Verified fvm via the shared capability — the same hash-pinned fetch
-      # every other job uses, not an unverified `pub global activate`. This job
-      # holds the pub.dev credentials, so it must not acquire tooling through a
-      # weaker path than the rest of the pipeline.
+      # Use the verified fvm capability, not an unverified pub global activate:
+      # this job holds the pub.dev credentials, so don't weaken its tooling path.
       - uses: ./.github/actions/capabilities/fvm
       - name: Build filtered changelog
         env:
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -267,14 +267,11 @@ jobs:
           ref: '${{ needs.discover.outputs.tag }}'
           fetch-depth: 0
           persist-credentials: false
-      - uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260 # v1.7.2
-      - name: Install fvm + project SDK
-        # Pin the fvm TOOL (the Flutter SDK is pinned in .fvmrc). Version
-        # lives in tool/versions.env so it never floats on @latest.
-        run: |
-          source tool/versions.env
-          dart pub global activate fvm "$FVM_VERSION"
-          fvm install
+      # Verified fvm via the shared capability — the same hash-pinned fetch
+      # every other job uses, not an unverified `pub global activate`. This job
+      # holds the pub.dev credentials, so it must not acquire tooling through a
+      # weaker path than the rest of the pipeline.
+      - uses: ./.github/actions/capabilities/fvm
       - name: Build filtered changelog
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/full-test.yml
+++ b/.github/workflows/full-test.yml
@@ -55,6 +55,7 @@ jobs:
   inputs:
     name: Resolve inputs
     runs-on: ubuntu-24.04
+    timeout-minutes: 5
     permissions: {}
     outputs:
       runner: ${{ steps.resolve.outputs.runner }}
@@ -184,6 +185,7 @@ jobs:
     if: always()
     needs: [inputs, test]
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 5
     permissions: {}
     steps:
       - name: Evaluate
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -3,6 +3,12 @@
 # makes GitHub match it: creates missing, recolours/redescribes changed, and
 # deletes any label not listed. Never edit labels in the UI; the next run
 # reverts the drift. The sync is authoritative, so labels.json must be complete.
+#
+# Description style: a terse phrase, sentence case, no trailing period; a comma
+# adds at most one substantive qualifier; no prefixes, no semicolons; noun
+# phrase unless the label's whole job is an action. Every one ends with an
+# applier tag: (auto) when a bot applies or manages it in the normal flow,
+# (manual) when you apply it.
 name: Labels
 
 on:
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -30,6 +30,7 @@ jobs:
   inputs:
     name: Resolve inputs
     runs-on: ubuntu-24.04
+    timeout-minutes: 5
     outputs:
       runner: ${{ steps.resolve.outputs.runner }}
     steps:
@@ -45,6 +46,7 @@ jobs:
     name: Conventional Commit
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 5
     steps:
       - name: Check PR title
         shell: bash
@@ -72,6 +74,7 @@ jobs:
     name: Promotion Chain
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 5
     steps:
       - name: Validate promotion path
         shell: bash
@@ -103,6 +106,7 @@ jobs:
     name: Changelog Versions
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 10
     permissions:
       contents: read
     steps:
@@ -133,6 +137,7 @@ jobs:
     name: Workflow lint
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 15
     permissions:
       contents: read  # checkout only; both tools read the workflow files
     steps:
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -165,6 +165,7 @@ jobs:
     name: Pin availability
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 15
     permissions:
       contents: read  # checkout only; HEADs the pinned asset URLs
     steps:
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -156,3 +156,21 @@ jobs:
       - name: shell lint (shellcheck + bash 3.2 portability)
         shell: bash
         run: bash tool/lint_shell.sh
+
+  # Runs on PR activity (not just the daily cron, which GitHub disables after
+  # ~60 idle days): HEADs every pinned asset so a pruned pin fails the PR that
+  # sits on it, before it fail-closes a build. Fails on a definitive 404/410,
+  # warns on transient.
+  pin-availability:
+    name: Pin availability
+    needs: inputs
+    runs-on: ${{ needs.inputs.outputs.runner }}
+    permissions:
+      contents: read  # checkout only; HEADs the pinned asset URLs
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - name: Pinned assets still reachable
+        shell: bash
+        run: bash tool/ci/upgrade.sh check-availability
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -47,18 +47,25 @@ jobs:
     needs: inputs
     runs-on: ${{ needs.inputs.outputs.runner }}
     timeout-minutes: 5
+    permissions:
+      contents: read  # checkout only; reads tool/commit-types.txt
     steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
       - name: Check PR title
         shell: bash
         env:
           TITLE: ${{ github.event.pull_request.title }}
         run: |
-          if ! grep -qE '^(feat|fix|docs|chore|ci|test|refactor|perf|build|deps)(\(.+\))?!?: .+' <<< "$TITLE"; then
+          # Same type list as the commit-msg hook (tool/commit-types.txt).
+          types=$(paste -sd'|' tool/commit-types.txt)
+          if ! grep -qE "^(${types})(\(.+\))?!?: .+" <<< "$TITLE"; then
             echo "::error::PR title must follow Conventional Commits format."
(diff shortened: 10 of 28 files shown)


slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — single-line fix that restores `gh` target-repo discovery without a checkout.

✅ **Checked**
- `GH_REPO: ${{ github.repository }}` uses a trusted context value, no injection vector
- The comment explains the "why" directly — matches the repo's existing convention
- No other step in this workflow has the same gap (every other `gh` call has a prior checkout)
- Workflow triggers are `push` and `workflow_dispatch` — no `pull_request_target` concerns
- `run:` line is clean; all event/input values kept in `env:`, per repo convention

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — release-critical GH_REPO fix is correct, all 12 supply-chain items properly applied or declined with recorded rationale

✅ **Checked**

- `create-release.yml:223` — GH_REPO set before `gh release upload`; the step runs before checkout (no .git), so the explicit repo name prevents "not a git repo" failure. Correct.
- `create-release.yml:274` — publish job now uses `.github/actions/capabilities/fvm` instead of unverified `pub global activate`. The publish job holds pub.dev credentials, so this closes a tool-acquisition gap.
- `tool/ci/upgrade.sh:67-81` — `set_kv` shape-validates every value before writing versions.env (sha256: 64 lowercase hex; version: `[A-Za-z0-9._+-]+`; unknown key shape: rejected). A malformed upstream string carrying a quote cannot be written, therefore cannot be executed on the next `source`.
- `tool/ci/upgrade.sh:45-46,285-288` — `blocked` counter accumulates failed asset fetches; exits 1 with `::error::` instead of the old silent exit 0.
- `tool/ci/upgrade.sh:99-148` — `verify-pinned` mode re-hashes every already-pinned asset, comparing to stored sha256. A hash change under a fixed version (upstream repoint) fails the run. The pins job runs this before `apply`, so a repoint is caught before any bump touches versions.env.
- `tool/fetch_verified.sh:26-33` — fail-closed allowlist: normalize (lowercase, strip whitespace), then accept only `^[0-9a-f]{64}$`. One rule catches empty, TODO, truncated, uppercase, stray-space — strictly better than the old blocklist of known placeholders.
- `tool/fetch_verified.sh:35-40` — `--speed-limit 2048 --speed-time 30` aborts stalls without killing slow-but-moving downloads. `--max-time` raised to 1800 as a generous backstop against a hung pipe.
- `.github/actions/capabilities/rust/action.yml:99-115` — portable awk replaces GNU-only `sed` insert. `SCCACHE_BIN` rides `ENVIRON` (not `awk -v`), so a path with backslashes stays literal. Tempfile + mv avoids partial writes.
- `.github/workflows/upgrade-check.yml` — single `upgrade` job correctly split into `pins` (security-sensitive, runs verify-pinned first, bumps + re-hashes) and `lockfiles` (routine, resolved against current dev SDK). Hash changes review undiluted by lock churn.
- `docs/ARCHITECTURE.md:619-624` — principle 3 now accurately describes jq-in-bash (via `json_get`) and Dart-not-from-bash (hook/build.dart reads build.json in Dart).
- `.github/labels.json` — two new labels (`upgrade-pins`, `upgrade-locks`) with proper auto/manual applier tags, consistent description style.
- All workflows: no untrusted data interpolated into `run:` blocks, no `pull_request_target`, checkout always on trusted refs, SHA-pinned actions throughout.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — release fix is correct, supply-chain hardening is well-reasoned, no blocking issues.

1. `.github/actions/capabilities/rust/action.yml:108` — The awk `ENVIRON["sk_w"]` pattern is solid, but the intermediate file write `"$CFG" > "$CFG.tmp" && mv "$CFG.tmp" "$CFG"` will truncate `$CFG` on awk failure before the `mv` runs (the redirect opens the file for writing before awk starts). Consider `awk '...' "$CFG" > "$CFG.tmp" && mv "$CFG.tmp" "$CFG"` — same logic, but the redirect happens after the awk process starts, so a failed awk launch leaves the original file intact. In practice this is unlikely to matter here (awk is present on all runners), but it matches the `set_kv` pattern in `upgrade.sh` which already uses the safer `tmp="$3.tmp"; ... > "$tmp"; then mv`.

✅ **Checked**
- `create-release.yml`: GH_REPO fix is correct — the upload step has no checkout, `gh release upload` needs `GH_REPO` to resolve the target.
- `create-release.yml` publish job: switching from `dart pub global activate fvm` to the shared `capabilities/fvm` action closes the tool-acquisition gap. The comment about "this job holds pub.dev credentials" is accurate.
- `upgrade.sh` set_kv shape validation: `*SHA256*` requires 64 lowercase hex, `*VERSION*` requires `[A-Za-z0-9._+-]+`. Both patterns correctly reject embedded quotes, backticks, and dollar signs — a malformed upstream string cannot be written to the sourced `versions.env`. The inner `[._+-]` bracket is POSIX-safe: `-` is literal because it appears last before `]`.
- `upgrade.sh` verify-pinned mode: re-hashes the currently-pinned assets before any bump, compares to stored sha256, and fails loudly on mismatch. Covers the repoint attack class that version-compare and shape-check are both blind to. Wired correctly as `upgrade-check.yml` pins job step 1.
- `upgrade.sh` blocked tracking: asset-fetch failures during `apply` now emit `::error::` annotations and exit nonzero instead of silently succeeding. The final `[ "$blocked" -ne 0 ] && exit 1` is reachable (the verify-pinned early-exit only fires for that mode, not for `check`/`apply`).
- `upgrade.sh` Flutter version gate: the `^[0-9]+\.[0-9]+\.[0-9]+$` check against `flutter_latest` runs before the value reaches the `sed s/.../` replacement — prevents the GNU sed `e` flag injection path noted in the comment.
- `fetch_verified.sh` sha256 allowlist: normalization (lowercase + whitespace strip) then exact `^[0-9a-f]{64}$` match. Fails closed on empty, TODO, truncated, uppercase, whitespace-padded — one rule covers all cases the old blocklist enumerated individually.
- `fetch_verified.sh` curl stall detection: `--speed-limit 2048 --speed-time 30` aborts on genuine stalls while letting slow-but-moving downloads finish. `--max-time 1800` is a generous backstop. Both are correct for large assets.
- `upgrade-check.yml` pins/lockfiles split: separate PRs with distinct branches (`chore/pins`, `chore/lockfiles`) and labels (`upgrade-pins`, `upgrade-locks`). No race — independent jobs. The repoint alarm (verify-pinned) only runs on the pins job, which is the correct scope.
- `labels.json`: new `upgrade-pins` and `upgrade-locks` labels match what `upgrade-check.yml` creates. Description style is consistent (noun phrase, no trailing period, applier tag).
- `ARCHITECTURE.md` principle 3: now correctly states bash uses jq via `lib.sh`'s `json_get`, and the build hook uses Dart. Matches reality.
- `lint_shell.sh`: both declined-lint comments are accurate and well-reasoned — the dangling-path false-positive problem is real, and the doc-vs-code tool-ban grep cannot distinguish scoped-true from global-false.
- Workflow security: no `pull_request_target`, no fork-triggerable workflows with write permissions, no untrusted `github.event.*` values interpolated into `run:` blocks. All checkouts use `persist-credentials: false`.
- Local gates report green: shellcheck, lint_shell (bash 3.2 + BSD/GNU), zizmor `--persona=auditor` at zero, actionlint.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — every injection sink is gated, every download is fail-closed, and the release fix unblocks the pipeline.

✅ **Checked**
- **Release fix** (`create-release.yml:218-223`): `GH_REPO` on the upload step — confirmed the step runs before any checkout, so gh has no `.git` to infer the target. `GH_REPO` names it explicitly, matching the pattern `release.sh` already uses.
- **set_kv validation** (`upgrade.sh:67-77`): `*_VERSION` regex `^[A-Za-z0-9._+-]+$` and `*_SHA256` regex `^[0-9a-f]{64}$` gate every write into `versions.env`. A crafted upstream string cannot reach the sourced file.
- **flutter_latest semver gate** (`upgrade.sh:148-153`): runs `^[0-9]+\.[0-9]+\.[0-9]+$` on the raw upstream value before it reaches `sed`. Stable Flutter versions are `X.Y.Z` — exact match, zero false positives, and the last raw sink is closed.
- **fetch_verified 64-hex allowlist** (`fetch_verified.sh:26-32`): `tr 'A-Z' 'a-z' | tr -d '[:space:]'` normalization + `^[0-9a-f]{64}$` regex rejects every non-hex shape (empty, TODO, truncated, uppercase, whitespace) in one rule. The old blocklist missed, e.g., a 63-char string. This catches it.
- **Portable cargo wrapper write** (`rust/action.yml:99-107`): awk uses `ENVIRON["sk_w"]`, not `awk -v` which interprets backslashes. The insert-under-`[build]` logic handles the pre-existing header case on both GNU and BSD awk.
- **verify-pinned** (`upgrade.sh:95-141`): `_vp` re-hashes every pinned asset; 404 → "prune, not tamper" (return 0); hash mismatch → `::error::` + fail. Runs before any bump, so a repoint is caught before it can be promoted into a new pin version.
- **Blocked bump nonzero exit** (`upgrade.sh:45-46, 280-283`): `blocked=1` set on every asset-fetch failure; final `exit 1` with `::error::`. Was silent `exit 0`.
- **Publish via verified fvm** (`create-release.yml:267-272`): `dart-lang/setup-dart` + unverified `pub global activate fvm` replaced with `uses: ./.github/actions/capabilities/fvm`, which downloads fvm via `fetch_verified.sh` with a pinned sha256. The publish job holds pub.dev credentials — this was the weakest tooling path.
- **ARCHITECTURE.md principle 3** (`ARCHITECTURE.md:619-624`): confirmed `json_get` in `lib.sh` uses `jq -er` and `ensure_jq` guarantees jq is present. The old text claimed "pure sed/grep" / "no jq" — both false.
- **Workflow security**: all modified workflows use `schedule`, `workflow_dispatch`, or `push` on specific branches. No `pull_request_target`. No fork-triggerable secrets.
- **Upgrade radar split** (`upgrade-check.yml`): pins and lockfiles are independent jobs pushing to separate branches, creating separate PRs with dedicated labels. No race condition, no hash-changes-buried-in-lock-churn.
- **Label schema** (`labels.json`, `labels.yml`): descriptions now end with `(auto)`/`(manual)` applier tags; style convention documented. New `upgrade-pins`/`upgrade-locks` labels for the split radar.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — release fix correct, supply-chain hardening comprehensive and well-reasoned

✅ **Checked**

- `create-release.yml`: `GH_REPO: ${{ github.repository }}` on the upload step correctly fixes the no-`.git` inference failure — the step runs before checkout, so `gh` had no repo to target
- `fetch_verified.sh`: fail-closed 64-hex allowlist replaces the placeholder blocklist; uppercase/spaces normalized before the anchor regex, so `"TODO"`, `"unknown"`, truncation, and mixed-case all hit the refusal. Stall-bound download (`--speed-limit 2048 --speed-time 30`) catches a hung pipe without capping large-asset transfers
- `upgrade.sh` `set_kv`: shape validation gates every write to the sourced `versions.env` — hex-only for `*SHA256*`, `[A-Za-z0-9._+-]+` for `*VERSION*`, reject on unknown key shape. Value travels via `ENVIRON`, not `awk -v` / sed replacement string
- `upgrade.sh` verify-pinned: re-hashes already-pinned assets before the bump step; alarm-first ordering prevents a bump on top of a repointed asset. 404 during verify is correctly treated as prune (not tamper) and returns 0
- `upgrade.sh` flutter version gate: `^[0-9]+\.[0-9]+\.[0-9]+$` validation on `flutter_latest` before sed interpolation — this is the one path where the value bypasses `set_kv`, so its own gate is the right call
- `upgrade.sh` blocked counter: each asset-download failure sets `blocked=1` and the script exits nonzero at the end; was a silent `exit 0`
- `upgrade-check.yml`: pins/lockfiles split into separate jobs hitting separate branches (`chore/pins`, `chore/lockfiles`) — hash changes review on their own, undiluted by lock churn. Both jobs use `persist-credentials: false` and wire `gh auth setup-git`
- `create-release.yml` publish job: uses `.github/actions/capabilities/fvm` (hash-verified download) instead of unverified `pub global activate fvm` — correct for the job that holds `PUB_CREDENTIALS`
- `rust/action.yml`: portable awk (`ENVIRON["sk_w"]`, not `awk -v`) replaces GNU-only `sed` insert; works on macOS
- `labels.json`: `upgrade-pins` and `upgrade-locks` labels added, descriptions follow the new (auto)/(manual) convention documented in `labels.yml`
- `docs/ARCHITECTURE.md` principle 3: scoped accurately — bash uses jq for JSON, the Dart hook reads `build.json` in Dart, bash never shells out to dart or python
- No `pull_request_target`, no untrusted interpolation into `run:`, secrets in `env:` only, SHA-pinned actions throughout

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — thorough supply-chain hardening with no blocking issues

1. `.github/workflows/pr-lint.yml:176` — the new `pin-availability` job has no `timeout-minutes`. In the worst case (all 22 URLs timing out with `--connect-timeout 10`, `--retry 2`, `--retry-delay 2`), the job could run for ~11 minutes before the 6-hour default timeout kills it. Add `timeout-minutes: 15` to bound it explicitly.

✅ **Checked**
- `GH_REPO: ${{ github.repository }}` in create-release.yml — `github.repository` is a trusted context value, not user input
- `pull_request` trigger on the new `pin-availability` job, not `pull_request_target` — fork code runs, but with read-only `contents: read` and no secrets
- All GitHub Actions refs are SHA-pinned — no rolling `@main` or `@v1` anywhere
- `set_kv` shape validation — `*SHA256*` gates to `^[0-9a-f]{64}$`, `*VERSION*` to `^[A-Za-z0-9._+-]+$`; unknown key patterns refuse. Values travel via `ENVIRON` (not `awk -v`), so no backslash/semicolon injection into the awk replacement
- Flutter semver gate — `^[0-9]+\.[0-9]+\.[0-9]+$` on `flutter_latest` before it reaches `sed`; covers GNU sed's `e` command (RCE), the last raw sink
- `fetch_verified.sh` fail-closed — one 64-hex allowlist replaces the old blocklist; normalizes case and whitespace before checking. `--speed-limit 2048 --speed-time 30` bounds stall, `--max-time 1800` is the backstop
- `verify-pinned` — HEADs each URL first (404/410 = pruned, 200 with changed hash = repoint), warns on transient codes. Runs in `upgrade-check.yml` before bumping
- `check-availability` — HEAD-only existence check, called from `pr-lint.yml` on every PR so a stale pin fails before the daily cron goes idle
- `blocked` counter — all four bump blocks set `blocked=1` on asset fetch failure; the script exits nonzero at the end, so a blocked bump no longer passes silently
- Upgrade radar split — pins PR and lockfiles PR are now independent; a hash change cannot be buried in lockfile churn. `verify-pinned` runs before any bump
- Publish job now routes through `.github/actions/capabilities/fvm` — hash-verified fvm download instead of unverified `pub global activate`. The publish job holds `PUB_CREDENTIALS`, so this closes a real tooling-path gap
- Cargo config write uses portable awk + ENVIRON — no longer GNU-only `sed`, no backslash interpretation via `-v`. Temp file cleanup on awk failure (`|| rm -f "$CFG.tmp"`)
- `versions.env` single-writer lint — catches redirects, tee, and in-place sed targeting `versions.env`; constructs named in prose so the rule cannot match itself
- Labels — descriptions follow the documented schema, consistent (auto)/(manual) applier tags, new `upgrade-pins`/`upgrade-locks` labels match the split workflow
- ARCHITECTURE principle 3 corrected — bash uses jq (`json_get`), the build hook reads `build.json` in Dart
- Shell portability — all new code uses bash 3.2-compatible constructs (no associative arrays, no case modification, no `mapfile`); awk, `printf`, `grep -qE` all portable across BSD/GNU

The hardening here is well-scoped and grounded in specific attack surfaces — every gate has a named sink, and the new lint rule enforces the single-writer invariant mechanically rather than by discipline. Nice work.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — supply-chain hardening is sound, release unblock is correct, and every gate passed

1. `tool/ci/upgrade.sh:54` — `_fetch`'s `--max-time 30` is re-used by `sha256_of()` for full binary downloads (Chrome is ~340MB). This was pre-existing; the new `blocked=1` now surfaces a timeout as a hard failure rather than a silent pass. Consider giving `sha256_of` its own curl call with a longer timeout so the daily bump does not fail on a slow link.

✅ **Checked**
- `fetch_verified.sh` fail-closed sha256 allowlist — `[0-9a-f]{64}` is exact, blocks empty/TODO/truncated/uppercase/whitespace in one rule
- `upgrade.sh` set_kv shape validation — SHA256 and VERSION both gated before writing the sourced file; blocked bumps exit nonzero
- `upgrade.sh` semver gate on flutter_latest — `^[0-9]+\.[0-9]+\.[0-9]+$` closes the sed injection sink on the one path that bypasses set_kv
- `upgrade.sh` verify-pinned — HEADs each asset before re-hashing; 404/410 fails, transient warns, 200 with changed hash fails
- `upgrade.sh` check-availability — HEAD-only, fails on definitive 404/410, runs on every PR so a pruned pin is caught even when the cron is disabled
- `upgrade-check.yml` pins/lockfiles split — pin bumps review on their own, verify-pinned runs first, lockfiles refresh independently
- `pr-lint.yml` pin-availability job — `pull_request` trigger (not target), `contents: read`, `persist-credentials: false`
- `create-release.yml` GH_REPO fix — `github.repository` in env, not run; trusted context value
- `create-release.yml` publish job — uses the hash-pinned fvm capability instead of an unverified `pub global activate`
- `rust/action.yml` cargo config — awk with ENVIRON replaces GNU-only sed insert; portable across macOS
- `lint_shell.sh` single-writer lint — greps for redirect/tee/sed -i targeting versions.env; tested for false-positives
- `labels.json` descriptions — consistent schema with (auto)/(manual) applier tags
- `ARCHITECTURE.md` principle 3 — corrected to reflect jq-backed json_get
- Shell lint, zizmor auditor zero, actionlint clean per PR body
- No `pull_request_target` in any changed workflow file
- `persist-credentials: false` on all checkout steps that lack push auth

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — release fix correct, supply-chain hardening solid.

1. `tool/lint_shell.sh:164-165` — `vw_files` rebuilds the same file list as `scan_files` (lines 45-52), running both `git ls-files` invocations a second time. Reuse `scan_files` directly instead.

2. `tool/ci/upgrade.sh:97` — `asset_urls()` reads sourced-env values set at script start (line 33). If it is ever called after a `set_kv` bump, the in-process env vars are stale. Currently it is only called from `verify_pinned` and `check_availability`, which run before any `set_kv` call, so it is correct — but a comment noting that ordering dependency would prevent a future regression.

✅ **Checked**

- `GH_REPO` fix correctly resolves the `gh release upload` "not a git repo" error
- `set_kv` shape validation gates every write to `versions.env` (64-hex for SHA256 keys, `[A-Za-z0-9._+-]+` for version keys, rejects unknown keys)
- `fetch_verified.sh` switched from a blocklist ("", "TODO", "unknown", "none") to a positive allowlist — only exactly 64 lowercase hex chars pass. Correctly fail-closed
- `verify-pinned` re-hashes already-pinned assets before the bump to catch same-version repoints; fails on 404/410 (pruned pin) and hash mismatch
- `check-availability` HEAD-only (no download or hash) runs on every PR via `pr-lint.yml`, catching a pruned pin even when the daily cron is disabled after 60 idle days
- `blocked=1` counter ensures a pin bump that fails its asset download exits nonzero instead of passing silently
- awk `ENVIRON` path for cargo config replaces GNU-only `sed -i.bak /a` insert — portable across macOS BSD sed and GNU sed
- Publish job now uses the hash-pinned `.github/actions/capabilities/fvm` instead of an unverified `dart pub global activate fvm`
- Upgrade radar split: pin bumps (`upgrade-pins` label + `chore/pins` branch) are their own reviewed PR, separate from lockfile churn (`upgrade-locks` label + `chore/lockfiles` branch)
- Single-writer lint (`lint_shell.sh` section 5) detects any redirect, tee, or `sed -i` writing to `versions.env` outside `set_kv`
- `--speed-limit 2048 --speed-time 30` bounds download stall without killing large assets on slow links; `--max-time 1800` is the hung-pipe backstop
- Flutter version gated to exact semver (`^[0-9]+\.[0-9]+\.[0-9]+$`) before it reaches sed — a separate sink from set_kv
- All workflow permissions are appropriately scoped; no injection vectors from user-controlled inputs in run blocks
- `timeout-minutes` added to all jobs across all workflows — consistent stall protection
- ARCHITECTURE principle 3 corrected: bash uses jq (via lib.sh), Dart reads JSON in the build hook
- Label descriptions standardized with (auto)/(manual) applier tags; two new labels `upgrade-pins` and `upgrade-locks` in labels.json

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — release pipeline unblocked, supply-chain hardening is sound, no blocking issues across all 14 files

✅ **Checked**

- **Release fix:** `GH_REPO: ${{ github.repository }}` in `create-release.yml:227` — correct. The upload step runs before its checkout, so `gh` has no `.git` to infer from. This is the regression from #113 that this unblocks.
- **Verified publish:** `create-release.yml:276` replaces `dart-lang/setup-dart` + `dart pub global activate fvm` with the hash-pinned `./.github/actions/capabilities/fvm`. The publish commands downstream use `fvm dart`, which the capability action provides. The old path was unverified (no hash pin on `pub global activate`); this job holds pub.dev credentials, so the hardening matters.
- **set_kv injection hardening:** `tool/ci/upgrade.sh:68-77` — each value is now shape-checked before writing to the sourced `versions.env`. VERSION must match `[A-Za-z0-9._+-]+`, SHA256 must be exactly 64 lowercase hex. A malformed upstream string is rejected at write time, not later at execution time.
- **Flutter version gate:** `tool/ci/upgrade.sh` line ~237 — `flutter_latest` is validated against `^[0-9]+\.[0-9]+\.[0-9]+$` before it reaches the `sed` replacement. GNU sed's `e` command makes this a real injection surface; the gate prevents a crafted upstream value from reaching it.
- **fetch_verified fail-closed allowlist:** `tool/fetch_verified.sh:27-31` — the old check was a blocklist (`""|"TODO"|"unknown"|"none"`). The new check accepts only exactly 64 lowercase hex, normalizing case and whitespace first. Any other shape (truncated, uppercase, empty, placeholder) is refused. Also: `--speed-limit 2048 --speed-time 30` bounds on stall, not total time, so large assets on slow links still finish.
- **verify-pinned tamper detection:** `tool/ci/upgrade.sh` — new `verify-pinned` mode re-hashes all currently-pinned assets. A hash change without a version change is a same-version repoint and fails the run before any bump happens. The pins job runs this first (`upgrade-check.yml` line 92).
- **check-availability PR-time pin check:** `pr-lint.yml:161-176` — HEAD requests every pinned asset on every PR. A definitive 404/410 fails the PR (catches a pruned pin even when the daily cron is disabled after 60 idle days). `contents: read`, `persist-credentials: false`, no secrets. The `pull_request` trigger (not `pull_request_target`) means PR branch code runs, but with no write perms or secrets. The `http_status` function uses curl without auth headers — no token leak.
- **Blocked exit:** `tool/ci/upgrade.sh:310-313` — when an asset download fails during bump, `blocked=1` causes the script to exit nonzero instead of passing silently. Previously the old code echoed a plain message and continued.
- **Portable awk:** `.github/actions/capabilities/rust/action.yml:104-108` — replaces GNU `sed -i.bak` with awk using `ENVIRON`, not `awk -v` (avoids backslash interpretation). Works on macOS and Linux.
- **Pin/lockfile split:** `upgrade-check.yml` — security-sensitive pin bumps get their own draft PR (`chore/pins`, label `upgrade-pins`), reviewed undiluted by lockfile churn. Lockfiles refresh separately (`chore/lockfiles`, label `upgrade-locks`). Each job checks out `ref: dev` (trusted code), triggered by `schedule`/`workflow_dispatch` (trusted contexts).
- **Single-writer lint:** `tool/lint_shell.sh:157-176` — new lint check fails if any shell script or workflow YAML writes to `versions.env` outside `set_kv`. The grep patterns match redirects, `tee`, and `sed -i`. The comment says constructs are named in prose (not pasted), so the lint doesn't match itself.
- **Architecture doc fix:** `docs/ARCHITECTURE.md:619-624` — principle 3 corrected: bash uses jq (via `json_get` in `lib.sh`), not hand-rolled `sed`/`grep`. The build hook (`hook/build.dart`) reads JSON in Dart.
- **Labels consistency:** `labels.json` — every label now has an `(auto)` or `(manual)` applier tag, consistent with the style doc added to `labels.yml`. Two new labels (`upgrade-pins`, `upgrade-locks`) with distinct colors.
- **Timeout-minutes on all jobs:** every job across all 5 workflows (ci, create-release, full-test, pr-lint, upgrade-check) now has a `timeout-minutes`. No job can hang indefinitely.
- **Workflow triggers safe:** `pr-lint.yml` uses `pull_request` (not `pull_request_target`). `upgrade-check.yml` and `create-release.yml` use `schedule`/`workflow_dispatch` and `push` respectively — all trusted contexts. No fork-triggerable secret exposure.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — release fix, CI hardening, and supply-chain checks are all consistent and correct.

## Changes reviewed (push 2)

**Release-critical fix** — `create-release.yml`: `GH_REPO: ${{ github.repository }}` on the `gh release upload` step, which runs before any checkout. Without it, `gh` has no `.git` to infer the repo and exits with "not a git repo." Clean, minimal, correct.

**Publish hardening** — same workflow: the publish job now uses the hash-verified FVM capability instead of `dart-lang/setup-dart` + an unverified `pub global activate fvm`. Since this job holds pub.dev credentials, the downgrade from verified to unverified tooling was worth closing.

**Supply-chain checks** — `upgrade.sh` gains two new modes:
- `verify-pinned`: re-downloads and re-hashes every currently pinned asset on the daily radar, fails on 404/410 (pruned) or 200-with-changed-hash (repoint). Runs *before* the bump.
- `check-availability`: HEAD-only existence check on every PR. Fails on definitive 404/410, warns on transient. Closes the gap where a disabled daily cron could hide a pruned pin until the build broke.

**Injection hardening** — `set_kv` in `upgrade.sh` now shape-validates every value before writing `versions.env` (64-hex for sha256 keys, `[A-Za-z0-9._+-]+` for version keys). `lint_shell.sh` adds a check that flags any write to `versions.env` not going through `set_kv` — a redirect, tee, or in-place sed would slip an unvalidated value into a sourced file. The Flutter SDK path (which bypasses `set_kv` for sed) gets a separate exact-semver gate before interpolation.

**Fail-loud** — blocked pin bumps (asset fetch failed) now set `blocked=1` and the script exits nonzero, where before it passed silently with a plain `echo`.

**Portable cargo config** — `sed -i.bak '/^\[build\]/a ...'` replaced with awk using `ENVIRON`, not `awk -v`. The prior form was GNU-only and broke on macOS.

**fetch_verified** — switched from a blocklist (`""|"TODO"|"unknown"|"none"`) to an allowlist (exactly 64 hex chars, fail-closed). Stalls are now bounded by transfer rate (`--speed-limit 2048 --speed-time 30`) with a 30-minute backstop, not a flat 3-minute wall clock that could kill a large download on a slow link.

**Upgrade radar split** — `upgrade-check.yml` refactored from one `upgrade` job to `pins` + `lockfiles` jobs, each opening separate draft PRs (`chore/pins` with label `upgrade-pins`, `chore/lockfiles` with label `upgrade-locks`). Labels, docs, and SECURITY.md updated accordingly.

**Workflow timeouts** — all five workflows now have `timeout-minutes` on every job (ranging 5–45 minutes per the PR description). All checkouts use `persist-credentials: false` and SHA-pinned actions. No `pull_request_target` triggers. No user-controlled input interpolated into `run:` blocks.

✅ **Checked**
- All 16 pinned asset URLs in `asset_urls()` match the bump-code URLs — the single-source pattern is consistent
- `check-availability` only fails on 404/410; transients produce warnings, so a CDN blip doesn't block PRs
- `fetch_verified.sh` normalization (`tr 'A-Z' 'a-z' | tr -d '[:space:]'`) is safe — the allowlist regex `^[0-9a-f]{64}$` still enforces 64 hex chars
- The `pin-availability` job sources `tool/versions.env` correctly via `upgrade.sh`'s own `source` on line 35
- bash 3.2 portability: the new `lint_shell.sh` check uses only process substitution and array append, both supported in 3.2
- Labels sync: `labels.json` adds `upgrade-pins` and `upgrade-locks` with `(auto)` tags and consistent descriptions
- Docs: ARCHITECTURE principle 3, UPDATING, and SECURITY all match the new behavior

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — supply-chain hardening is thorough and correct; release fix unblocks the pipeline.

✅ **Checked**
- `create-release.yml`: `GH_REPO` env var on the upload step correctly names the repo for gh when no checkout precedes it — the release fix
- `create-release.yml` publish job: switched from `dart-lang/setup-dart` + `pub global activate fvm` to the hash-verified fvm capability, closing the weakest toolchain link in the job that holds pub.dev credentials
- `fetch_verified.sh`: fail-closed allowlist (`^[0-9a-f]{64}$`) replaces the placeholder blocklist, with case/whitespace normalization before the check. Stall-bound download (`--speed-limit 2048 --speed-time 30`) with a 30-minute backstop
- `upgrade.sh` `set_kv`: shape-gates every value before writing the sourced `versions.env` — sha256 pin to 64 hex, version to `[A-Za-z0-9._+-]+`. Flutter version gets a separate, stricter `^[0-9]+\.[0-9]+\.[0-9]+$` gate before it reaches sed (the last raw sink)
- `upgrade.sh` `blocked` counter: a failed asset download now emits `::error::` and exits nonzero at the end instead of passing silently
- `upgrade.sh` `verify-pinned` / `check-availability`: HEAD + re-hash for repoint detection (daily), HEAD-only for prune detection (per-PR in pr-lint.yml). `asset_urls()` is the single source of truth for both
- `upgrade-check.yml` split: security-sensitive pin bumps on their own `chore/pins` PR, lockfiles on `chore/lockfiles`. `verify-pinned` runs first on the pins job before any bump
- `release.sh`: `valid_semver()` gates changelog versions before they flow into sed/awk/file paths. `stamp_asset_hashes` uses process substitution (not `| while` subshell) so a bad digest can `return` and fail the function, and asserts `sha256:` algorithm prefix. `false_secrets` insertion handles the missing-key case. `cmd_check_versions` validates the release-candidate heading shape
- `lint_shell.sh` check 5: enforces that only `set_kv` writes `versions.env` — a sourced file, so any other writer is an unvalidated injection path
- `rust/action.yml`: `CARGO_HTTP_MULTIPLEXING=false` for curl 16 errors; awk (portable) replaces GNU-only sed for the cargo config insert, path via ENVIRON not `-v`
- Workflow hardening: all 19 previously unbounded jobs now have `timeout-minutes` (5–45). `persist-credentials: false` and per-job permissions were already uniform. No `pull_request_target` in any workflow. `github.event.pull_request.title` flows into an env var, used as grep data via heredoc — no injection
- Docs (ARCHITECTURE, SECURITY, UPDATING) are consistent with the code changes

No blocking issues found. The hardening is well-reasoned and each change is documented in its commit message and inline comment.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — supply-chain and workflow hardening is thorough and every change checks out

✅ **Checked**

- **Release fix.** `create-release.yml`:228 — `GH_REPO: ${{ github.repository }}` lets `gh release upload` find the repo without a checkout. Correct.
- **set_kv shape gate.** `tool/ci/upgrade.sh`:73-79 — validates each value as sha256 (64 hex) or version (`[A-Za-z0-9._+-]+`) before writing the sourced `versions.env`. Unknown key shapes are rejected. The `tr A-Z a-z` + `tr -d '[:space:]'` normalization in `fetch_verified.sh`:28 happens before the allowlist regex, so uppercase or whitespace-padded hashes are normalized then checked.
- **Flutter semver gate.** `tool/ci/upgrade.sh`:182-185 — gates `flutter_latest` to exact `^[0-9]+\.[0-9]+\.[0-9]+$` before it reaches the `sed` on `.fvmrc`. This is the right gate for Flutter stable releases, and it is the only raw sink (the other `*_latest` vars reach only curl URLs and set_kv).
- **Fail-closed allowlist.** `tool/fetch_verified.sh`:29-32 — `^[0-9a-f]{64}$` fails closed on empty, TODO, truncated, uppercase, and whitespace. The old blocklist only caught the specific strings it named.
- **Stall-bound download.** `tool/fetch_verified.sh`:36-38 — `--speed-limit 2048 --speed-time 30` bounds on progress rather than a flat `--max-time`, with `--max-time 1800` as backstop. Correct for large assets on slow links.
- **verify-pinned + check-availability.** `tool/ci/upgrade.sh`:133-169 — re-hashes pinned assets to catch same-version repoints (200 + hash mismatch = error); HEAD-only check on every PR catches pruned pins (404/410 = error, transient = warning). Both flow through `http_status()` which follows redirects and always exits 0 so `$(...)` never trips `set -e`.
- **Blocked bumps fail loud.** `tool/ci/upgrade.sh`:47,213-214,312-315 — `blocked` counter set when an asset fetch fails; the script exits nonzero at the end. Previously the run passed silently.
- **Portable awk cargo config.** `.github/actions/capabilities/rust/action.yml`:107-111 — awk replaces the GNU-only `sed -i.bak '/^\[build\]/a ...'` that broke on macOS. The sccache binary path passes via `ENVIRON`, not `awk -v`, so backslashes in a Windows path stay literal. The `|| rm -f "$CFG.tmp"` on line 111 cleans up on failure.
- **split upgrade PRs.** `.github/workflows/upgrade-check.yml` — pins (`chore/pins`) and lockfiles (`chore/lockfiles`) now ride separate PRs. The pins job runs `verify-pinned` first, then bumps. Both jobs use SHA-pinned checkout with `persist-credentials: false`.
- **Verified publish.** `create-release.yml`:278 — the publish job now uses `.github/actions/capabilities/fvm` instead of `pub global activate fvm`. Correct: this job holds pub.dev credentials and should not weaken its tooling path.
- **Timeout hardening.** 19 jobs across 5 workflows now bounded: 5 min for orchestration, 10-20 min for CI checks, 30-45 min for heavy builds. The 6-hour default fallback is gone.
- **Workflow security.** No `pull_request_target` triggers. `github.event.*` values all routed through `env:` before reaching `run:` blocks. All action refs are SHA-pinned. `persist-credentials: false` on all checkouts. Per-job permissions match each job's needs — no over-granting.
- **pr-lint pin-availability.** `pr-lint.yml`:176-189 — runs `check-availability` on every PR open/edit/sync/reopen. `contents: read` only, no secrets. Forks safe.
- **githooks.** `commit-msg` now reads types from `tool/commit-types.txt` (single source of truth, shared with pr-lint). `pre-commit` uses `--no-index` for git check-ignore (load-bearing for force-added files) and NUL-delimited paths end to end.
- **Makefile build-native.** `Makefile`:120-126 — greps for build-hook failure markers after the `tee`-piped dart invocation, so a compile failure does not pass silently behind `|| true`.
- **Shell lint.** `tool/lint_shell.sh`:163-173 — new check that only `set_kv` writes `versions.env`; flags any redirect, tee, or in-place sed targeting the file.
- **release.sh hardening.** `valid_semver()` gates version strings before they reach seds/awk/file paths. `stamp_asset_hashes` validates digest algorithm (sha256 only) and uses process substitution instead of a `| while` subshell. The `false_secrets` block now handles the missing-key case. Semver validation runs in `cmd_check_versions` so a malformed version fails the PR.
- **Docs.** ARCHITECTURE principle 3 corrected (bash uses jq via `json_get`, not pure sed/grep). SECURITY updated to mention verify-pinned/check-availability. UPDATING documents the split upgrade flow.
- **Local gates.** Shellcheck, lint_shell (bash 3.2 + BSD/GNU), zizmor auditor at zero, actionlint — all green per PR body.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — every change is a hardening or correctness fix; no regressions introduced.

1. `tool/ci/upgrade.sh:172` — `verify_pinned && exit 0; exit 1` is correct but terse. If `verify_pinned` fails, the error messages are already on stderr from inside the function, but adding `echo "verify-pinned: FAILED" >&2` before the exit would make the failure more visible in the CI log at a glance.
2. `tool/ci/upgrade.sh:47` — `blocked` is documented as "nonzero at the end instead of passing silently", but its usage spans ~300 lines. A comment at each `blocked=1` assignment (near lines 213, 253, 275, 304) noting "sets blocked for the fail-loud exit at end of script" would help future readers connect the dots.

✅ **Checked**
- **Workflow security**: no `pull_request_target` triggers, all untrusted inputs (`github.event.pull_request.title`, `github.head_ref`) in `env:` blocks, no token exposure to fork PRs.
- **Release fix**: `GH_REPO` in `create-release.yml` correctly sets the repo for `gh release upload` when no checkout precedes the step.
- **Shell injection hardening**: `set_kv` shape-gates values (64-hex sha or version regex) before writing `versions.env`; `valid_semver` gates Flutter versions before they reach `sed`; `fetch_verified.sh` now fails closed on a 64-hex allowlist instead of a blocklist placeholder.
- **Supply-chain pinning**: `verify-pinned` re-hashes existing pins to catch upstream repoints before bumping; `check-availability` HEADs pins on every PR so a pruned pin fails the PR rather than a build. `blocked` counter ensures a failed asset fetch during `apply` exits nonzero — previously it passed silently.
- **Pin/lockfile split**: security-sensitive pin bumps (`chore/pins`) and lockfile refreshes (`chore/lockfiles`) now get separate reviewed PRs.
- **Timeout-minutes**: all 19 previously unbounded jobs now have explicit timeouts (5–45).
- **Checkout hardening**: SHA-pinned checkouts with `persist-credentials: false` throughout.
- **Correctness fixes**: `build-native` now detects build-hook failures via output markers (was `|| true` swallowing all errors); the gitignore pre-commit hook now uses `--no-index` so force-added files are properly detected; the conventional-commit type list is now a single source of truth in `tool/commit-types.txt`.
- **CODEOWNERS**: `/vendor/` now owner-gated — a submodule pointer bump pulls in arbitrary `build.rs` code that cargo runs at compile time.
- **Portable cargo config**: `awk` replaces a GNU-only `sed` insert that broke on macOS; value passed via `ENVIRON`, not `awk -v`, avoiding backslash interpretation.
- **Verified publish**: the publish job now uses the hash-pinned fvm capability instead of an unverified `pub global activate`.
- **Release script hardening**: `stamp_asset_hashes` validates digest algorithm; `cmd_discover` handles missing `false_secrets:` section; `cmd_gate` handles empty `BEFORE` explicitly; semver validation catches malformed changelog headings at PR time.
- **Docs**: `SECURITY.md`, `ARCHITECTURE.md`, `UPDATING.md` updated to reflect the new verify-pinned/check-availability flow and the pins/lockfiles split.

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — thorough hardening pass, release fix is correct, every gate is sound

✅ **Checked**
- Release fix: `GH_REPO` on `upload-assets` (line 228) correctly names the repo so `gh release upload` works before any checkout; publish job now uses the hash-pinned fvm capability instead of `pub global activate` (line 278)
- Supply-chain: `set_kv` shape-validates every value (64-hex sha or version chars) before writing the sourced `versions.env`; `verify-pinned` re-hashes all pinned assets and fails on a repoint; `check-availability` HEADs every pin on each PR so a disabled cron cannot hide a pruned asset; `fetch_verified` fails closed on any non-64-hex hash; blocked pin bumps exit nonzero
- Workflow security: timeout-minutes on all 19 jobs across five workflows; all checkouts SHA-pinned with `persist-credentials:false`; `github.event.pull_request.title` placed in `env:` (`pr-lint.yml` line 59), not interpolated into a `run:` block; no `pull_request_target` events; per-job permissions already uniform
- Hooks: `--no-index` on `git check-ignore` correctly applies gitignore rules regardless of index state; NUL-safe path handling with `-z`; single source of truth for commit types (`tool/commit-types.txt`) shared by both the commit-msg hook and pr-lint
- Build: `build-native` now detects compile failures via output markers instead of swallowing the exit code; test-guards use `io-exempt:` comment markers with reasons rather than positional `grep -v` paths
- CODEOWNERS: `/vendor/` owner-gated for submodule pointer bumps; the comment correctly explains why per-file patterns inside submodule repos are deliberately absent
- Docs: ARCHITECTURE principle 3 (jq/Dart), SECURITY self-computed-pin note (verify-pinned/check-availability), UPDATING upgrade flow (pins/lockfiles split) — all consistent with the code
- Dev gates: `lint_shell.sh` check 5 enforces `set_kv` as the sole writer of `versions.env`; `tool/commit-types.txt` eliminates the duplicated type list; local gates (shellcheck, lint_shell, zizmor auditor, actionlint) report green

PR #122, opened 2026-06-24
chore: promote dev to prod (2.1.1)

chaudharydeepanshu:
Promotes `dev` to `prod`, cutting the `2.1.1` release.

Contents (one squashed PR):
- #121 — unblock the release pipeline (release push auth + `gh release upload` repo inference) plus a supply-chain, workflow, and dev-tooling-gate hardening pass.

The previously-cut `v2.1.1` GitHub release/tag were deleted; this promotion re-cuts it cleanly. pub.dev never received 2.1.1 (latest there is 2.1.0), so there's no published version to retract.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -21,12 +21,15 @@ if printf '%s\n' "$msg" | grep -qE '^Merge '; then
   exit 1
 fi
 
-if ! printf '%s\n' "$msg" | grep -qE '^(feat|fix|docs|chore|ci|test|refactor|perf|build|deps)(\(.+\))?!?: .+'; then
+# Single source of truth for the type list: tool/commit-types.txt (pr-lint.yml
+# validates PR titles against the same file). cwd is the repo root during hooks.
+types=$(paste -sd'|' tool/commit-types.txt)
+if ! printf '%s\n' "$msg" | grep -qE "^(${types})(\(.+\))?!?: .+"; then
   echo "ERROR: Commit message must follow Conventional Commits format."
   echo ""
   echo "  <type>[optional scope]: <description>"
   echo ""
-  echo "  Types: feat fix docs chore ci test refactor perf build deps"
+  printf '  Types: %s\n' "$(paste -sd' ' tool/commit-types.txt)"
   echo "  Examples:"
   echo "    feat: add watermark support"
   echo "    fix(web): handle empty PDF edge case"
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -2,9 +2,13 @@
 # Reject commits that add files matching .gitignore patterns.
 # Catches accidental `git add -f` of ignored files.
 
-bad_files=$(git diff --cached --name-only --diff-filter=A | while read f; do
-  git check-ignore --quiet "$f" 2>/dev/null && echo "$f"
-done)
+# --no-index is load-bearing: a force-added file is already in the index, and
+# without it git check-ignore treats an indexed path as tracked and reports
+# nothing, so the hook would silently never fire. NUL end to end
+# (core.quotepath=false stops octal-escaping, -z stops newline-splitting) keeps
+# paths with spaces or newlines intact; --stdin avoids a non-POSIX read -d '' loop.
+bad_files=$(git -c core.quotepath=false diff --cached --name-only --diff-filter=A -z \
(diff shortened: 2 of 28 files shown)


slopfairy:
posted my review below 👇

slopfairy:
PR #122 is already merged (closed at 07:53 UTC). The `synchronize` event fired post-merge, so there is no review to submit.

I read the full diff — 28 files, +486/-138 — and spot-checked the security-sensitive paths:

- `pr-lint.yml` triggers on `pull_request:` (not `pull_request_target`), so the new `checkout` step and the `${{ github.event.pull_request.title }}` through `env:` → heredoc is safe.
- `create-release.yml`: `GH_REPO` fix for `gh release upload` + `.github/actions/capabilities/fvm` replaces the unverified `pub global activate fvm` path.
- `fetch_verified.sh`: sha256 validation upgraded from a case-esac denylist to a strict 64-hex regex gate — fully fail-closed.
- `upgrade.sh`: `set_kv` validates value shapes before writing to the sourced `versions.env`; `verify-pinned` re-hashes on the daily radar; `check-availability` HEADs on every PR; flutter version validated against semver before sed interpolation.
- `release.sh`: `valid_semver()` gates version strings; `stamp_asset_hashes` now validates digest algorithms; handles missing `false_secrets` in `pubspec.yaml`.
- `lint_shell.sh`: new `versions.env` single-writer lint check.
- `upgrade-check.yml`: split into `pins` + `lockfiles` jobs so hash changes never sit in lockfile churn.
- `CODEOWNERS`: `/vendor/` submodule pointers now require owner review.
- All workflows got `timeout-minutes` on every job.

No issues found. The changeset is a clean hardening pass.

slopfairy (approved):
> [!TIP]
> Approved — supply-chain, workflow, and dev-tooling hardening pass; safe to promote to prod.

✅ **Checked**

- **Security hardening**: `fetch_verified.sh` now fail-closes on any non-64-hex hash (instead of enumerating placeholder strings); `upgrade.sh` `set_kv` validates value shape before writing to a sourced file; `release.sh` asserts digest algorithm (`sha256:` only) and validates semver before it hits a sed/awk program; `lint_shell.sh` now enforces that only `set_kv` writes to `versions.env`. All correct.
- **Supply-chain**: `upgrade-check.yml` split into separate `pins` and `lockfiles` jobs, with a `verify-pinned` pre-check that re-hashes already-pinned assets before bumping (repoint alarm). `pr-lint.yml` adds a `pin-availability` job that HEADs every pinned asset on PR activity, so a disabled daily cron can't hide a pruned pin. `CODEOWNERS` gates `/vendor/` to the owner. `create-release.yml` publish job switches to the verified `.github/actions/capabilities/fvm` instead of `dart pub global activate`.
- **Workflow security**: No `pull_request_target` triggers anywhere. All untrusted data (`github.event.pull_request.title`, `github.base_ref`, `github.head_ref`) routed through `env:` before use in `run:` blocks — correct pattern. All checkouts use `persist-credentials: false`. Top-level `permissions: {}` with per-job least-privilege grants.
- **Bug fixes**: `.githooks/pre-commit` `--no-index` fix — `git check-ignore` previously treated force-added files as tracked and silently skipped them. `Makefile` `build-native` no longer masks build failures with `|| true`. `rust/action.yml` sccache config insertion uses `awk` instead of GNU `sed a` (fixes macOS). `release.sh` `false_secrets` insertion handles the missing-section case and verifies the result.
- **DRY**: `tool/commit-types.txt` is the single source of truth for conventional commit types, read by both `.githooks/commit-msg` and `pr-lint.yml`. `upgrade.sh` `asset_urls()` centralises the pin table (one place to keep in sync with the bump blocks).
- **CI reliability**: `timeout-minutes` on every job across all four workflows; `CARGO_HTTP_MULTIPLEXING=false` for crates.io; `GH_REPO` env for `gh release upload` when no checkout exists; `--speed-limit` replaces `--max-time` in `fetch_verified.sh` so large assets on slow links finish.
- **Test infrastructure**: `io-exempt:` comment-based guard replaces a hardcoded exclusion list in the Makefile. All seven `dart:io` test files carry the new marker. Clearer, self-documenting, no false negatives.

This promotion follows the linear dev→prod chain. Everything in it was reviewed in #121 and is now cleanly assembled for the 2.1.1 release.

PR #126, opened 2026-06-24
ci: verify the dependency floor compiles (analyze on oldest deps)

chaudharydeepanshu:
Closes the dependency half of "works for some, breaks for some" for pdf_manipulator as a published package.

**Gap:** CI only built on the newest resolution, so the wide lower bounds (`package_config: >=2.1.0 <4.0.0` and friends) were claimed but never exercised. A consumer resolving the oldest in-range deps got an untested package.

**Fix:** `make analyze-floor` downgrades to the oldest in-range deps and analyzes the shipped code (`lib bin hook`). Static only: no fixtures, no native build, no submodules. Tests are excluded since a consumer never runs them. Snapshots and restores the lock so a local run leaves the tree clean. The CI gate now requires it, so a future bump that needs newer deps than the constraints advertise turns red instead of shipping.

**Verified:** dogfooded locally, the shipped code analyzes clean on its floor, so the `>=3.10.0` floor (forced by `hooks`/`code_assets`) and the wide constraints are honest.

Local gates: actionlint clean, zizmor `--persona=auditor` zero (no new ignores).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,24 @@ jobs:
         with:
           make-target: analyze
 
+  # Analyze the shipped code (lib/bin/hook) against the OLDEST in-range deps, so
+  # the wide lower bounds are honest, not just the newest a fresh build resolves.
+  # Static analysis only: no fixtures, no native build, no submodules needed.
+  analyze-floor:
+    name: Analyze (floor deps)
+    needs: [inputs, changes]
+    if: needs.changes.outputs.code == 'true'
+    runs-on: ${{ needs.inputs.outputs.runner }}
+    timeout-minutes: 15
+    permissions:
+      contents: read  # checkout + static analyze, read-only
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with: { persist-credentials: false }
+      - uses: ./.github/actions/make-target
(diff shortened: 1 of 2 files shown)


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — catches dependency floor regressions before they ship, clean across the board

✅ **Checked**
- `Makefile:analyze-floor` — error handling is correct: `downgrade` failure or `analyze` failure both captured in `rc`, lockfile restored unconditionally via `mv` + `pub get`, exit code preserved. Follows the established multi-line recipe pattern (`\` continuations with shared shell state) used by `build-native`, `test-guards`, and others.
- `bin/` exists (`setup.dart`) — `dart analyze lib bin hook` will resolve all three directories.
- CI job security — `pull_request:` trigger (not `pull_request_target`), `contents: read` permissions, `persist-credentials: false`, pinned `actions/checkout` SHA. No untrusted `github.event.*` values in `run:` blocks.
- Gate integration — `analyze-floor` added to `needs` and `FLOOR` env wired into the evaluation script. Edge cases all handled: skipped jobs (doc-only passes), failed jobs (gate fails), missing output from a failed `changes` (caught by the `CHANGES != success` guard first).
- No `submodules: recursive` on the `analyze-floor` checkout — correct, static Dart analysis doesn't need the vendored Rust crates.

PR #142, opened 2026-07-04
build: adopt the shared platforms gate + shell lint

chaudharydeepanshu:
Adopts the shared platform gate + shell lint from whuppi/ci@v1.0.7, retiring pdf's local copies.

- **`tool/platforms_gate.sh`**: stamped canonical gate, replaces `tool/platforms.sh` (deleted). Same pinned-pana 6-platform verdict, now shared so device_io and pdf can't drift. `PANA_VERSION` still comes from this repo's `tool/versions.env`.
- **`tool/lint_shell.sh`**: now the shared self-contained copy, no longer sourcing the build helpers.
- **`tool/lib.sh` → `tool/build_lib.sh`**: renamed now that lint_shell no longer drags it along; the 5 build scripts that source it are swept. Plus a standalone `make format` target.
- **Makefile**: `platforms` → the shared gate.
- **Pins** bumped v1.0.6 → v1.0.7 across workflows **and the vendored make-target action** (the fork that silently drifted to v1.0.1 before #140 — now caught by the version-consistency guard from #141).
- **Docs**: the CI/CD sections that re-explained shared whuppi/ci mechanics trimmed to cross-refs; kept only pdf's native rust/wasm layer.

Verified locally: `make lint-shell` clean, `make platforms` OK (pana 0.23.14, all six).

PR #141, opened 2026-07-04
ci: align dependabot/labels/upgrade-body with the workspace shape

chaudharydeepanshu:
Aligns pdf's dependabot/labels/labeler with the workspace shape, fixes the pdf#139 auto-body over-claim, and adds a fork-drift guard.

- **dependabot**: adds the missing `/example` pub entry — its example-app deps were never being bumped — and adopts the canonical shape now byte-identical to device_io's.
- **labeler + labels**: `Makefile` + `tool/**` added to the `ci` rule, a `docs` rule + label added (both match device_io; the `web`/`upstream` rules stay pdf-specific).
- **upgrade-check.yml**: the pins-PR body hardcoded "binaryen + pana … recomputed sha256s" even when only pana moved (the pdf#139 report). Now generic — points at the diff for which versions moved.
- **pr-lint**: new check fails the PR if `whuppi/ci` refs across `.github` span versions. The vendored `make-target` fork drifted to v1.0.1 while workflows were v1.0.4, and dependabot's `whuppi-ci` group doesn't reach inside `.github/actions/` — this catches a recurrence.

Config/workflow only. YAML+JSON verified, `make lint-shell` green. Part of the device_io↔pdf↔ci consistency streamlining.

_Note: the cosmetic workflow renames (pr-lint→pr-checks, create-release→release) are held for a separate review — they touch the release pipeline + branch protection._

PR #143, opened 2026-07-04
ci: collapse the make-target fork into a thin wrapper

chaudharydeepanshu:
Collapses pdf's vendored `make-target` fork — the last item from the consistency audit, and the one I'd flagged as "needs a ci feature." It turned out **not** to: ci's make-target already has the `xcode-key-extra` input the fork needs, so this is a pure pdf refactor.

**Fork → thin wrapper.** The fork duplicated ~100 lines of whuppi/ci's generic orchestration just to interleave rust/wasm capabilities — and that duplication is what drifted (its ~13 whuppi/ci refs bumped separately from the workflows). Now it provisions the three pdf-specific capabilities (rust, wasm-cache, wasm-build) locally, then **delegates** the generic orchestration + make run to `whuppi/ci/actions/make-target`. **1 whuppi/ci ref instead of 13**; the generic logic can't drift because it's no longer copied.

**Behavior-preserving — verified against the capabilities + matrix:**
- rust/wasm-cache/wasm-build are all fvm-independent (compile_rust.sh is pure cargo), so running them before the shared action's fvm is safe.
- rust's `SUBMODULE_PDF_OXIDE` (`$GITHUB_ENV`) reaches the shared xcode-cache via `xcode-key-extra` — the exact cache key the fork built by hand.
- **no matrix job combines `free-disk` with `wasm-build`**, so wasm-build never needs disk reclaimed first (the one ordering delegation can't provide — checked every row).

> **Stacked on #142** (branched off it to avoid a fork-file conflict) — merge #142 first. **Added `ready-to-test`** so the full matrix exercises the wasm-build / emulator / xcode paths the light gate never touches — this is the real verification for a make-target change.


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — fork collapsed to a clean thin wrapper; two stale doc references flagged below

1. `.githooks/commit-msg:24` — comment references `pr-lint.yml`, which is now `pr-checks.yml`
2. `docs/ARCHITECTURE.md:782` — comment references `tool/lib.sh`, which is now `tool/build_lib.sh`

✅ **Checked**
- **action.yml** — local capabilities (rust, wasm-cache, wasm-build) run before the shared action. Ordering is safe: `compile_rust.sh` is pure cargo, and `SUBMODULE_PDF_OXIDE` reaches xcode-cache via `xcode-key-extra`. Every generic input (`free-disk`, `java`, `chrome`, etc.) forwarded to `whuppi/ci/actions/make-target@v1.0.7` without omission.
- **Version bumps** — all 13 workflow refs bumped from `v1.0.6` to `v1.0.7` consistently across `auto-close.yml`, `debug-ssh.yml`, `full-test.yml`, `labels.yml`, `pr-checks.yml`, `release.yml`, `retry.yml`, `triage.yml`, `upgrade-check.yml`.
- **Workflow renames** — `pr-lint.yml` → `pr-checks.yml`, `create-release.yml` → `release.yml`. All references in AGENTS.md, docs/ARCHITECTURE.md, and docs/UPDATING.md updated.
- **Shell script refactor** — `lib.sh` → `build_lib.sh` rename consistent across `analyze.sh`, `check_alignment.sh`, `compile_rust.sh`, `lint_shell.sh`, `tool/ci/upgrade.sh`. No stale `source` paths remain.
- **`lint_shell.sh`** — self-contained now; no `lib.sh` dependency. `git rev-parse --show-toplevel` validates it's in a repo without relying on a computed ROOT. The Makefile invokes it from the repo root (`make lint-shell`), so the relative `git ls-files` calls resolve correctly.
- **`platforms_gate.sh`** — canonical copy from whuppi/ci, stamped in. Parameterized via `EXPECTED_PLATFORMS` and `PANA_VERSION` from the local `versions.env`. Same pana gate, same rsync snapshot, now shared verbatim across repos.
- **Makefile** — `platforms` target updated to call `platforms_gate.sh`. New `format` target added (standalone format pass, each package from its own root). `.PHONY` updated.
- **`full-test.yml`** — matrix still passes `wasm-cache`/`wasm-build`/`emulator`/`simulator` to `make-target` exactly as before. Verified against the PR body's claim: no row combines `free-disk` with `wasm-build`.
- **Security** — `triage.yml` still uses `pull_request_target` delegating to the shared hardened workflow, unchanged except the version bump. Workflow permissions remain least-privilege.

chaudharydeepanshu:
Marking draft — the thin-wrapper collapse **consistently regresses `int: Android (macos-intel)`** and shouldn't ship as-is.

**Confirmed (not a flake):**
- This wrapper fails macos-intel Android **2/2** (initial run + a targeted re-run).
- The fork passes macos-intel Android **2/2** (#142's full-test runs).
- The *identical* wrapper passes **all 24 other matrix jobs** — every wasm-build (ubuntu/macos/win), every xcode (iOS/macOS), and Android on **ubuntu, windows, and macos-arm64**.

So the delegation itself is sound; the failure is specific to the slow Intel-macOS Android runner. The symptom is a `pdf_smoke_test.dart` load failure after the Android build.

**Likely mechanism (not fully pinned):** the wrapper inverts the fork's `fvm → rust` order (a delegating wrapper structurally runs rust *before* it hands off to make-target's fvm). That ordering is load-bearing on slow runners in a way macos-arm64 tolerates but macos-intel doesn't. The `rust` capability's own comment already flags sccache-server survival across "the 25+ min gap (SDK download + emulator boot + Gradle)" as fragile — lengthening that gap is the kind of thing the wrapper does.

**Decision:** keep the fork. It works, and its *drift* — the only real problem — is already fixed (all refs v1.0.7) and guarded (the version-consistency check in #141). Collapsing it isn't worth shipping an Intel-runner regression. Pinning the exact interaction, or an `fvm`-first wrapper variant, could revisit this later, but the fork is the safe state today.

chaudharydeepanshu:
Closing — decision landed: keep the fork (it works; drift fixed + guarded), and harden the stamp discipline instead. The Rust-in-make-target rework revisits only when a 2nd Rust package exists, gated on a green macos-intel matrix. Full rationale in ci/docs/ARCHITECTURE.md.

PR #146, opened 2026-07-06
fix: dart2wasm web build; adopt whuppi/ci v2.0.0 dual-compiler gate

### LINKED ISSUE #145: pdf_manipulator 2.1.0 fails to compile with Flutter Web WASM / dart2wasm on Dart 3.11
### What happened?

  Hi! I’m trying to build a Flutter web app with WASM enabled and `pdf_manipulator 2.1.0` fails during `dart2wasm` compilation.

  ## Environment

  - Target: Flutter Web WASM
  - Command:

  ```bash
  flutter build web --wasm --release
 ```
  ## Error
 ```
  ../../.pub-cache/hosted/pub.dev/pdf_manipulator-2.1.0/lib/src/runtime/web/web_lane.dart:606:24:
  Error: The type 'Object?' is not exhaustively matched by the switch cases since it doesn't match 'Object()'.

        obj[k] = switch (v) {
                         ^
 ```
  Try adding a wildcard pattern or cases that match `Object()`.

  Relevant code appears to be:
 ```dart
  obj[k] = switch (v) {
    null => null,
    final String s => s.toJS,
    final int n => n.toJS,
    final bool b => b.toJS,
    final JSAny js => js,
  };
 ```
  ## Expected

  The package should compile for Flutter Web WASM / dart2wasm.

  With that patch applied locally, flutter build web --wasm --release succeeds.

### Steps to reproduce

1. use pdf_manipulator
2. compile with `flutter build web --wasm --release`

### Target (which platform?)

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [x] Web

### Package version

2.1.1

### Flutter version

3.41.9

### Did this work before?

None

chaudharydeepanshu:
## The fix (closes #145)
`lane.dart`'s `_post` switched over `Object?` with a `JSAny` arm. **dart2js treats `JSAny` as a catch-all (exhaustive); dart2wasm doesn't (non-exhaustive)** — so 2.1.0 fails a Flutter Web WASM / `dart2wasm` build while every CI check stayed green. Replaced the switch with `v.jsify()`, the canonical Dart→JS conversion that's identical on both compilers.

## Why CI missed it
Nothing in the toolchain actually compiles wasm: the analyzer + dart2js use the JS type model (switch looks exhaustive), the `invalid_runtime_check_with_js_interop_types` lint covers `is`/`as` but not switch patterns, and pana's wasm tag is an import heuristic. This adopts whuppi/ci **v2.0.0**, whose new `verify_web_gate.sh` compiles the example under **both** `flutter build web` and `--wasm` — the only honest check.

## Adoption (v2.0.0 is MAJOR — gates now require inputs)
- Pins v1.0.8 → v2.0.0; re-stamped all gates + manifest.
- Makefile: `make platforms` passes `EXPECTED_PLATFORMS`; `make verify-web` calls the new gate.
- `analyze.sh` requires DART/FLUTTER (no `${VAR:-…}` fallback).

Verified: `make analyze` / `platforms` / `verify-web` all green — both compilers, all 6 platforms.

### DIFF
--- a/.github/actions/make-target/action.yml
+++ b/.github/actions/make-target/action.yml
@@ -75,7 +75,7 @@ runs:
     # the native/wasm ones (rust, wasm-cache, wasm-build) are pdf-specific and
     # stay local. The shared capabilities were generalized FROM these, so the
     # behavior is identical — only their home moved.
-    - uses: whuppi/ci/actions/capabilities/fvm@v1.0.8
+    - uses: whuppi/ci/actions/capabilities/fvm@v2.0.0
 
     # ── On demand ──
     # Default on: most jobs build the native/wasm engine. Dart-only static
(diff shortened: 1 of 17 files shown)


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — correct fix for dart2wasm switch exhaustiveness, backed by a real dual-compiler gate that would have caught it.

✅ **Checked**
- `lib/src/runtime/web/lane.dart:620` — replacing the switch with `v.jsify()` is the right call. `jsify()` handles null, String, int, bool, and JSAny identically on both compilers. The old switch relied on `JSAny` as an erased catch-all, which dart2js accepts but dart2wasm rejects as non-exhaustive. The callers only pass basic Dart types (no JSAny flows through `_post`), so `jsify()` is a direct replacement.
- `tool/verify_web_gate.sh` — compiles the example under both `flutter build web` and `--wasm`, closing the gap that let #145 ship. Clean script, proper FLUTTER validation, well-commented rationale.
- `tool/analyze.sh`, `tool/analyze_core.sh`, `tool/platforms_gate.sh` — consistent hardening of DART/FLUTTER/EXPECTED_PLATFORMS from `${VAR:-fallback}` to `: "${VAR:?...}"`. Fails loud instead of guessing — the right pattern.
- `tool/stamped-files.txt` — correctly lists the 4 shared gates including the new verify_web_gate.sh.
- `Makefile` — `platforms` passes EXPECTED_PLATFORMS explicitly; `verify-web` delegates to the new dual-compiler gate.
- All 12 workflow/action version bumps (v1.0.8 → v2.0.0) are consistent. `pr-checks.yml`'s pin-availability check enforces one version.
- No security issues in workflows: no `pull_request_target`, no user-controlled values in `run:` blocks, actions are SHA-pinned.

PR #152, opened 2026-07-06
ci: debug-ssh reads the dispatch input directly — drop the resolve job

chaudharydeepanshu:
debug-ssh is dispatch-only and the runner input carries a default, so the Resolve-inputs job (copied from ci.yml, where push/PR triggers make it necessary) never does anything here. device_io's copy already reads inputs.runner directly; this aligns the template's older copy to that shape.

### DIFF
--- a/.github/workflows/debug-ssh.yml
+++ b/.github/workflows/debug-ssh.yml
@@ -1,7 +1,7 @@
 # ────────────────────────────────────────────────────────────────────
 # debug-ssh.yml — SSH into any runner for interactive debugging.
 #
-# Uses the debug-ssh composite action (bore tunnel, key-only auth).
+# Uses the shared debug-ssh composite action (bore tunnel, key-only auth).
 # Blocking mode — keeps the runner alive until /tmp/stop is touched.
 #
 # Usage:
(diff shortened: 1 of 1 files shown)


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — the resolve-inputs job was a no-op for a dispatch-only workflow with a required+defaulted input

The reasoning in the PR body is correct. `debug-ssh.yml` is triggered only by `workflow_dispatch`, and `runner` is `required: true` with `default: ubuntu-24.04`. That means `${{ inputs.runner }}` is always set — the `${OVERRIDE:-$DEFAULT}` shell expression never sees an empty value. The resolve job existed only because it was copied from `ci.yml`, where it is genuinely needed (push and `pull_request` triggers have no `inputs` context).

✅ **Checked**
- `.github/workflows/debug-ssh.yml:34` — `runs-on: ${{ inputs.runner }}` is safe: dispatch-only, choice input with fixed options, no untrusted interpolation into a `run:` block
- The removed `env.DEFAULT_RUNNER` had no other consumers — nothing left dangling
- Concurrency, permissions, timeout, and step ordering are unchanged

chaudharydeepanshu:
Going the other way for template uniformity: the resolve-inputs pattern stays everywhere; device_io gets the job added instead.

PR #156, opened 2026-07-07
fix: form fill/flatten non-ASCII text handling (ß, umlauts, CJK)

### LINKED ISSUE #155: Form fill/flatten mishandles non-ASCII text-string encodings (ß, umlauts, CJK)
Filling and flattening PDF forms breaks when the text involved isn't plain English. German is the easiest way to hit it: a form with a field called `Straße` (street), or a value like `München`, can fail three different ways. CJK and emoji values have their own failure mode on top.

## What it looks like

Say you have a fillable PDF with an address field and you do the usual thing:

```dart
await editor.setFormFieldValue('Straße und Hausnr', 'Königsallee 42');
await editor.flattenForms();
```

Depending on which program created the PDF, you get one of:

1. **"Form field not found"** — even though the field is right there and other tools fill it fine.
2. **Garbage in the flattened page** — `Königsallee` comes out as `K�nigsallee` or `KÃ¶nigsallee`.
3. **The value vanishes** — the flattened page shows an empty box where your text should be. This one even hits plain-English values on some PDFs.

Plain ASCII mostly works, which is why this went unnoticed: the characters ß, ä, ö, ü sit exactly in the range where the different text encodings PDFs use stop agreeing with each other.

## Why it happens

A PDF stores text like field names and values as raw bytes, and the format allows several encodings for those bytes. Different PDF creators pick different ones — Adobe tools use one (PDFDocEncoding), browser-based generators use another (UTF-16), LibreOffice writes plain UTF-8 (technically against the spec, but common). A correct reader has to detect which one it's looking at.

Our engine had **several independent pieces of code each decoding these bytes their own way** — and they disagreed:

1. **Fill-by-name** decoded field names strictly by the spec, so the UTF-8-style names LibreOffice writes turned into mojibake internally and your lookup missed them. Whether a PDF worked depended on which program made it.
2. **Flatten** read values back with the wrong decoder (turning `ß` into `�`), and then wrote the text into the page using UTF-8 bytes under a font that reads one byte per character — so every special character became two wrong ones (`ß` → `ÃŸ`). CJK and emoji drew nothing at all, because the fallback font that can render them wasn't compiled into the build.
3. **Fill then flatten** on fields without a prebuilt appearance lost the value entirely: the code that matches your new value back to the field on the page used yet another decoder for the field name, so for any non-ASCII name the match failed silently.

## Status

A 14-row test battery now covers the whole matrix (every name encoding × fill, save, flatten) and pinned all three bugs. The fix replaces the scattered decoders with one shared, tolerant one, writes page text one byte per character as the font expects, and bundles the fallback font for CJK/emoji. Landing together with the tests.

chaudharydeepanshu:
Closes #155.

Three bugs, one root cause: multiple ad-hoc text-string decoders disagreeing about the same bytes, plus an appearance bake that never encoded for its single-byte target font.

**Engine patches** (submodule, marked per the fork contract): one shared `decode_pdf_text_string` (UTF-16BE/LE BOM, UTF-8 BOM per ISO 32000-2, real-world raw-UTF-8 tolerance, PDFDocEncoding fallback) consumed by the annotation parser, form extractor, document decoder, and metadata reads; metadata writes via `Object::text_string`; appearance literals encoded one WinAnsi byte per char with the Windows-1252 specials mapped.

**Build**: `cjk-form-fonts` enabled in both lanes — flattening CJK/emoji values embeds the bundled fallback font.

**Tests**: new `form text encoding` battery, 14 rows over the producer-encoding matrix (PDFDocEncoding / UTF-16BE / raw UTF-8 / dart-pdf) with byte-exact handwritten fixtures. All rows failed before the engine patch; full native suite is 230/230 after. One documented nuance: the extractor's word-gap heuristic drops the space at a font-run boundary (CJK→Latin) — the space glyph is verifiably baked; asserted per-run with an in-place comment.

No changelog entry — cutting the next prerelease is a separate call.

### DIFF
--- a/test/fixtures/handwritten.dart
+++ b/test/fixtures/handwritten.dart
@@ -142,7 +142,19 @@ final Uint8List emptyBytes = Uint8List(0);
 /// Almost a PDF — has header but broken structure.
 final Uint8List brokenPdf = _build('%PDF-1.4\ngarbage after header\n');
 
-Uint8List _build(String content) => Uint8List.fromList(content.codeUnits);
+// Fixture strings are byte strings: each char's code unit IS the byte
+// (Latin-1 escapes like ß emit the single PDFDocEncoding byte 0xDF).
+// A char above U+00FF has no single byte — Uint8List.fromList would
+// silently truncate it to the low byte, corrupting the fixture. Spell
+// multi-byte sequences out as individual \u00XX escapes instead.
+Uint8List _build(String content) {
+  assert(
+    content.codeUnits.every((u) => u <= 0xFF),
+    'handwritten fixture contains a char above U+00FF — it would '
+    'silently truncate to its low byte',
+  );
+  return Uint8List.fromList(content.codeUnits);
+}
 
 final Uint8List bookmarkedPdf = Uint8List.fromList([
   0x25,
--- a/test/fixtures/handwritten.dart
+++ b/test/fixtures/handwritten.dart
@@ -9451,3 +9451,124 @@ final Uint8List testPkcs12 = Uint8List.fromList([
   0x27,
   0x10,
 ]);
+
+/// Minimal AcroForm whose field name, value, and document title are
+/// PDFDocEncoding literals (single Latin-1 bytes: 0xDF, 0xF6, 0xFC) --
+/// the ISO 32000-1 s7.9.2.2 non-UTF-16 branch. Adobe-InDesign-class
+/// producers ship forms with exactly this /T shape. The bytes are NOT
+/// valid UTF-8; a reader that decodes them as UTF-8 sees U+FFFD.
+/// Field "Prüfstraße und Hausnr", value "Königsallee 42, München",
+/// title "Straßen-Formular für Prüfung" -- all invented for this
+/// fixture.
+final Uint8List formPdfdocNamePdf = _build(
+  '%PDF-1.4\n'
+  '1 0 obj\n'
+  '<< /Type /Catalog /Pages 2 0 R /AcroForm << /Fields [4 0 R] /DR << /Font << /Helv 5 0 R >> >> /DA (/Helv 0 Tf 0 g) >> >>\n'
+  'endobj\n'
+  '2 0 obj\n'
+  '<< /Type /Pages /Kids [3 0 R] /Count 1 >>\n'
+  'endobj\n'
+  '3 0 obj\n'
+  '<< /Type /Page /Parent 2 0 R /MediaBox [0 0 595 842] /Annots [4 0 R] >>\n'
+  'endobj\n'
+  '4 0 obj\n'
+  '<< /Type /Annot /Subtype /Widget /FT /Tx /T (Pr\u00fcfstra\u00dfe und Hausnr) /V (K\u00f6nigsallee 42, M\u00fcnchen) /Rect [50 700 300 720] /DA (/Helv 12 Tf 0 g) /F 4 >>\n'
+  'endobj\n'
+  '5 0 obj\n'
+  '<< /Type /Font /Subtype /Type1 /BaseFont /Helvetica /Encoding /WinAnsiEncoding >>\n'
+  'endobj\n'
+  '6 0 obj\n'
+  '<< /Title (Stra\u00dfen-Formular f\u00fcr Pr\u00fcfung) >>\n'
+  'endobj\n'
+  'xref\n'
+  '0 7\n'
+  '0000000000 65535 f \n'
+  '0000000009 00000 n \n'
+  '0000000145 00000 n \n'
+  '0000000202 00000 n \n'
+  '0000000289 00000 n \n'
+  '0000000454 00000 n \n'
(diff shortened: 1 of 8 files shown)


slopfairy:
posted my review below 👇

test/fixtures/handwritten.dart
```diff
@@ -9451,3 +9451,124 @@ final Uint8List testPkcs12 = Uint8List.fromList([
   0x27,
   0x10,
 ]);
+
+/// Minimal AcroForm whose field name, value, and document title are
```
slopfairy, line 9467: The three handwritten fixtures are byte-correct for their respective encodings. The `_build` function at line 145 uses `content.codeUnits`, which maps Latin-1-range Dart Unicode escapes (e.g. `ü` → 0xFC) to single bytes — this is the correct PDFDocEncoding byte. The raw-UTF-8 fixture similarly maps `Ã¼` → 0xC3 0xBC, the actual UTF-8 sequence for ü. This only works because all characters stay in [U+0000, U+00FF], and the comments document the intent clearly.

slopfairy (dismissed):
> [!TIP]
> Approved — engine encoding fix + comprehensive test coverage; all Dart-side changes are correct.

✅ **Checked**
- `build.json` — `cjk-form-fonts` enabled on both native and wasm lanes, matching the PR's stated need for CJK/emoji flattening
- `test/fixtures/catalog.dart` — `form_fields_de` fixture correctly uses dart-pdf's `TextField` with non-ASCII name/value, and `_saveDoc` to serialize
- `test/fixtures/handwritten.dart` — all three handwritten fixtures (PDFDocEncoding, UTF-16BE+BOM, raw UTF-8) produce the intended bytes through `_build`; the Latin-1-range Dart escapes map to the correct single-byte values
- `test/ops/core/pdf_form_encoding_battery.dart` — 14 tests covering the full matrix: fill-by-name for every /T encoding, flatten-to-extract for Latin-1/CJK/special-char values, round-trip save-and-refill, and metadata decode; the CJK space-extraction nuance is clearly documented in-place
- Test runner registration — all four runners (native, web_atomics, web_jspi, web_opfs) register the battery
- CI workflows — `ci.yml` uses `pull_request` (safe for forks), all actions SHA-pinned; `triage.yml` uses `pull_request_target` with no PR checkout, pinned reusable workflow, least-privilege permissions
- Submodule bump — `vendor/pdf_oxide` moves from `98caaf5` to `27ff1b2`; the Rust changes live there per the fork contract, and the Dart side pins the exact rev
- No changelog entry — consistent with the repo's release-tool-managed versioning policy

test/fixtures/handwritten.dart
chaudharydeepanshu: Good catch on the invariant — made it mechanical in the follow-up commit: `_build` now asserts every code unit is ≤ 0xFF, so a future fixture char above U+00FF fails loudly at load instead of silently truncating to its low byte.

slopfairy: > @chaudharydeepanshu: "Good catch on the invariant — made it mechanical in the follow-up commit: `_build` now asserts every code unit is ≤ 0xFF…"

Confirmed — the assertion at line 151 is clean and will catch any above-U+00FF character at fixture load time rather than producing a truncated byte. Resolving this thread.

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — the `_build` invariant from the previous round is in, and the 14-test battery exhaustively covers the encoding matrix

**Changes since last review:**
- `test/fixtures/handwritten.dart:151` — added the `_build` assertion guarding against chars above U+00FF (addressed the prior review thread, now resolved)
- Three new handwritten fixtures: `formPdfdocNamePdf`, `formUtf16NamePdf`, `formUtf8NamePdf`
- One new catalog fixture: `form_fields_de` (dart-pdf German AcroForm)
- New battery at `test/ops/core/pdf_form_encoding_battery.dart` — 14 tests across fill-by-name, flatten→extract, round-trips, and metadata
- `cjk-form-fonts` enabled in both `build.json` build lanes
- Submodule pointer bumped to `27ff1b26`

✅ **Checked**
- `_build` assertion fires before `Uint8List.fromList`, catching any fixture author who uses a char above U+00FF — exactly what the previous review called for
- All three handwritten fixtures produce correct bytes: PDFDocEncoding literals (ß→0xDF, ü→0xFC), UTF-16BE hex with BOM, and raw UTF-8 (0xC3 0x9F, 0xC3 0xBC) — all pass the ≤ 0xFF assertion
- The battery imports match the generator naming convention (`fFormFieldsDe` / `fFormFieldsDeTruth`)
- All four test runners register the battery (native, web atomics, JSPI, OPFS)
- No changelog needed per the release process
- The CJK test's split assertion (CJK run separate from Latin run) is well-documented — the extractor's font-run boundary heuristic is a known limitation, and glyph fidelity is correctly verified per-run

PR #169, opened 2026-07-17
feat: trim — ship only the ops the app calls (#167)

### LINKED ISSUE #167: Ship only the ops the app calls
> [!NOTE]
> **Status:** shipped in #169. Trimming is in, sizes are measured and guarded. The compiler-driven `@RecordUse` lane is built and waiting on the upstream SDK experiment. Details in the last comment.

The engine is one compiled binary per platform, with every operation compiled into it. An app that uses only one operation still includes the code for every other operation. That makes the binary larger than it needs to be. It's most noticeable on web, where the binary loads over the network.

This is a universal problem, because Flutter doesn't tree-shake non-Dart code.

**Context**

The [roadmap](https://github.com/whuppi/pdf_manipulator/blob/dev/docs/CAPABILITY_ROADMAP.md) doc notes a rough plan. The plan is not final. Some of it depends on upstream Dart tooling that isn't ready yet (`@RecordUse` for instance methods, and FFI tree-shaking), and may not be ready soon. I plan to research the options further, and may use **temporary solutions** until an official path is available.

This is a large problem, so it will need time and research before I can work on it. I'm opening it for visibility, both for me and for anyone else with the same problem, and I do plan to tackle it.

chaudharydeepanshu:
Ships 2.2.0-dev.0: the trim system plus a smaller default binary for everyone.

- **Breaking:** flattening CJK/emoji form values no longer uses a bundled font (4.4 MB in every install). Register one once with `Pdf.registerFallbackFont`; unregistered values still save correctly — only the baked-in appearance falls back to the field's font.
- Trim (#167): `trim: auto` under `hooks: user_defines: pdf_manipulator:` scans the app and keeps what it can reach; `trim: {keep: [...]}` is the manual form; web adds `setup --trim`. Needs Rust (compiles once, cached). Wrong or missing pieces fail with a clear message, never a broken app.
- Default binary: dead engine surfaces, unused barcodes, and embedded fonts removed — native 28.7 MB → 21.1 MB, gzipped web download 11.3 MB → 7.2 MB. Fully trimmed: native 6.3 MB, web 1.9 MB on the wire.
- Engine updated — web consumers re-run `flutter pub run pdf_manipulator:setup --force web` after upgrading.

Full details: CHANGELOG.pre.md (2.2.0-dev.0) and the README's "The engine binary" section.

Closes #167


slopfairy:
posted my review below 👇

slopfairy:
posted my review below 👇

lib/src/runtime/router.dart
```diff
@@ -212,13 +215,55 @@ class Router implements PdfTransport {
...
+      return (await slot.lane.submit(job)).bytes;
+    } finally {
+      slot.activeJobs--;
+    }
+  }
+
   /// Least-loaded lane, spawning under the cap. (Pinned routing is
   /// resolved by the caller; an unknown handle falls through here —
   /// any lane can host the engine's "not found" error.)
   _LaneSlot _placeFor(Uint8List request) {
```
slopfairy, line 258: The `_preludes` list is iterated in `_placeFor` (line ~278) without synchronization. If `installPrelude` is called by one async context while `_placeFor` is spawning a lane in another, the concurrent modification could throw. In practice `registerFallbackFont` is called once during init, so this is unlikely to fire — but copying the list before iterating (`List.of(_preludes)`) costs nothing and closes the gap.

```suggestion
      for (final p in List.of(_preludes)) {
```

slopfairy (dismissed):
> [!TIP]
> Approved — well-designed trim system, conservative by construction, thoroughly tested.

1. `pubspec.yaml` — `analyzer: '>=7.0.0 <9.0.0'` is a direct dependency. The readme correctly says it is tooling-only (never exported from the library barrel, zero bytes in the compiled app), but every consumer still downloads it during `pub get`. The `<9.0.0` upper bound also forces the lock file to downgrade `analyzer` from 14.0.0 to 8.4.1 and `package_config` from 3.0.0 to 2.2.0 — those version swings are substantive and will constrain any other package in the consumer's pub solve that expects a newer analyzer. Consider whether the detector could live in a separate package or be refactored so this weight is opt-in.

2. `pubspec.yaml` — `record_use: ^0.6.0` as a direct dependency pulls experimental Dart SDK infrastructure into every consumer's pub solve. `record_use_shim.dart` is a clean no-op shim, but the dependency itself is dead weight for the 100% of consumers who will never touch the RecordUse lane. Consider moving it to a dev dependency and guarding the import behind a conditional — only `hook/link.dart` needs it at runtime, and only in the EXPERIMENTAL lane.

3. `lib/src/runtime/router.dart:277` — `for (final p in _preludes)` in `_placeFor` iterates the live list. The immediately preceding `for (final slot in List.of(_lanes))` in `installPrelude` (line 241) already copies defensively — the `_preludes` iteration should do the same. Copying the list (`List.of(_preludes)`) closes the gap for the cost of one allocation on lane spawn, which is already a cold path.

4. `analysis_options.yaml` — `experimental_member_use: ignore` is global. The justification (inline suppressions banned by tool/analyze.sh) is reasonable, but this silences ALL experimental member uses across the package, not just `@RecordUse`. If a future experimental API is used inadvertently elsewhere, this suppression hides it. Consider a scoped ignore if the analyzer supports it, or relax the inline-suppression ban for just this one case.

✅ **Checked**

- Trim design: fail-closed by contract. Any unresolvable file returns the full binary with a warning. Malformed `trim:` values fail the build printing the grammar. `featuresFor` only removes features whose capability is absent from the keep-set — internals (`icc`, `legacy-crypto`, `native-bridge`) survive unconditionally. Correct.
- `_preludes` replay on lane spawn: fire-and-forget with `unawaited` — a failed replay degrades that lane gracefully, never fails user jobs. The prelude rides the lane's FIFO so it executes before any user job on that lane. Correct.
- `web_assets/lane_worker.js`: replacing `__wbindgen_export` (renumbered on module changes) with `lane_alloc`/`lane_dealloc` (stable API surface) prevents memory corruption when the wasm module's export set shifts. The `__wbindgen_add_to_stack_pointer` call is kept because its name is stable. Correct.
- `lib/src/trim/detector.dart`: `_MemberFinder` visits `MethodInvocation`, `PropertyAccess`, and `PrefixedIdentifier` — covers direct calls (`pdf.sign(...)`), property access (`doc.render`), and prefixed usage (`pdf.doc.render`). `_record` filters by `package:pdf_manipulator/` library URI so calls from other packages do not pollute the keep-set. Correct.
- `lib/src/trim/capabilities.dart`: `requires` transitive expansion in `TrimConfig.keep()` and `featuresFor` dropping via set difference — `keep: [office]` correctly keeps `extract` by expansion and removes `rendering`, `signatures`, `pdfa` by subtraction. Parity test (`capabilities_test.dart`) reads `vendor/pdf_oxide/Cargo.toml` and asserts the Dart `requires` graph mirrors cargo's feature dependencies — drift between the two surfaces as a test failure. Correct.
- Breaking change (bundled CJK font removal): documented in CHANGELOG, README, and PR description. Migration path (`Pdf.registerFallbackFont`) is one call. Without a registered font, the fill still succeeds — only the baked-in appearance degrades, never data loss. Test battery (`pdf_form_encoding_battery.dart`) covers both the degraded and registered-font paths. Correct.
- `bin/setup.dart --trim`: resolves capabilities from the consuming app's source, fails closed on unresolved files, passes the feature override through `PDF_FEATURES_WASM` env var into the existing compile_rust.sh. Web consumers get a trimmed wasm compile; native consumers get the same trim through the build hook reading `userDefines['trim']`. Consistent.
- Vendor submodule bump: `vendor/pdf_oxide` moved to a new commit. The Makefile `test-rust` target now includes `public-api,cjk-form-fonts,pdfa` for test coverage on Rust side. `shake-audit` verifier keeps the trim guarantee durable across upstream rebases.

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — `trim_plan.dart` missing `break` causes manual keep-sets to silently fall through into auto-detection; `shake_audit.sh` is macOS-only; `link.dart` has a null-assert crash path

Five reviews dispatched across all 45 changed files (hooks, trim system, bridge/ops, CI/web assets, docs/tests). Below is the consolidated result.

## Summary

This is a substantial, well-structured PR. The trim architecture is clean: one shared compile path (`engine_compiler.dart`), two callers (`build.dart` + `link.dart`), a conservative detector (`detector.dart`), and a clear capability vocabulary (`capabilities.dart`). The record-use lane is properly dormant/EXPERIMENTAL. The bridge and ops changes for `registerFallbackFont` and `TrimRecord.op` are correct and consistent. The web asset shrink (289 exports removed via `public-api` feature gating) is impressive. Test coverage for the trim grammar is thorough.

Five blocking issues below: one correctness bug in the trim plan logic, two portability issues in `shake_audit.sh`, and two robustness issues in `link.dart`.

---

1. `lib/src/hook/trim_plan.dart:1919` — **Missing `break` after `case TrimMode.manual:`.** After setting `keep = config.keep`, execution falls through into `case TrimMode.auto:`, which runs the detector and overwrites the user's explicit keep-set. A `trim: {keep: [render]}` config silently takes the auto-detected result instead. Add `break;` after the `keep = config.keep;` line (see inline suggestion).

2. `tool/shake_audit.sh:6712,6718,6722,6755,6759` — macOS-only `stat -f%z` and `nm -gU` will fail on Linux. `make shake-audit` is a published Makefile target; if CI runs on ubuntu, the gate is broken. Also hardcodes `.dylib` extension in `dylib_for()` (line 6707) — on Linux the file is `.so`. Wrap in platform-aware helpers.

3. `tool/shake_audit.sh:6750,6790` — `python3` is assumed available but is not guaranteed on minimal CI images or stock macOS. Add a preflight check.

4. `hook/link.dart:1444` — `full.file!` is a forced null-assert on a nullable field. If `CodeAsset.file` is null (valid per the `code_assets` API), the link hook crashes. Add a null guard that ships the full binary (see inline suggestion).

5. `hook/link.dart:1452` — the cargo `targetDir` at `'cargo_target'` is shared across all code assets in the link loop. In a multi-architecture build, the second compile overwrites the first before `CodeAsset` is constructed. Use a per-asset target directory (see inline suggestion).

1. `hook/build.dart:994-995` — module-level `late` globals (`_constants`, `_plan`) have no null/reset guard. The old code had a lazy-load pattern.

2. `hook/build.dart:1234` — `_wasmFeaturesOverride` is a mutable global; pass as a parameter instead.

3. `lib/src/hook/engine_compiler.dart:1806` — `Platform.environment['PATH']!` null-assert, fragile in sandboxed environments.

4. `lib/src/trim/capabilities.dart:2435` — `featuresFor` is whitespace-sensitive. Trim feature names during split.

5. `bin/setup.dart:348-356` — `readWasmFeatures` re-parses `build.json` instead of using `BuildConstants.load()`.

6. `pubspec.lock` — `hooks 2.1.0 → 2.0.2`, `record_use 1.0.0 → 0.6.0`, `package_config 3.0.0 → 2.2.0`. Verify that `LinkInput.recordedUses` and the `Recordings`/`CallWithArguments`/`StringConstant` types exist at these versions.

7. `tool/shake_audit.sh:6707` — `dylib_for()` assumes `.dylib` (macOS-only). Linux `.so` and Windows `.dll` are needed for cross-platform CI.

✅ **Checked**

- Bridge/FFI boundary — `registerFallbackFont` flows correctly through the transport prelude mechanism. No null safety or isolate safety issues.
- Router prelude mechanism — snapshot iteration, fire-and-forget on new lanes, `activeJobs` management all correct.
- `bin/setup.dart` — `--trim` plumbing clean, fail-closed logic correct.
- `build.json` — feature list updated correctly: `cjk-form-fonts` removed, `pdfa,office,extract` added.
- `compile_rust.sh` — env-var override pattern (`${PDF_FEATURES_NATIVE:-$NATIVE_FEATURES}`) correct.
- `web_assets/pdf_oxide.js` — 289 exports removed via `public-api` feature gating, only lane ABI surface remains.
- `web_assets/lane_worker.js` — JSPI path uses stable `lane_alloc`/`lane_dealloc` instead of renumbered `__wbindgen_export_N`.
- `test/trim/capabilities_test.dart` — 13 grammar tests covering parsing, dependency expansion, errors, feature mapping, engine parity guard.
- `test/trim/trim_plan_test.dart` — plan matrix, RecordUse extraction, fail-closed paths.
- `test/ops/core/pdf_form_encoding_battery.dart` — CJK registration + degraded path tested.
- Changelog versioning — `2.2.0`/`2.2.0-dev.0` correct per independent-lane convention.
- README.md size figures — consistent with `PLAN_OP_TRIMMING.md` ledger.
- No `.github/workflows` changes — existing workflows pass the security audit.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — trim system, prelude/font architecture, and lane_worker.js stability fix are all solid

✅ **Checked**

- Trim grammar (`capabilities.dart`): three legal shapes, everything else fails loudly with the full grammar printed. Dependency expansion (`office` → `extract`) is correct and tested against the engine manifest.
- Trim detector (`detector.dart`): resolved-AST approach, fail-closed on any unanalyzable file. Correctly scoped to `package:pdf_manipulator/` elements only.
- Prelude system (`router.dart`): `installPrelude` replays on existing lanes synchronously, replays on future lanes fire-and-forget. The `_preludes` snapshot concern from the prior review is not a real issue in Dart — `ListIterator` captures `length` at creation, and no modification happens inside the for-loop body. Thread resolved.
- `shared_bridge.dart:registerFallbackFont`: encodes the font as a prelude, checks all existing-lane responses for engine errors, returns a `PdfTask` for error propagation.
- `lane_worker.js`: the switch from `__wbindgen_export` / `__wbindgen_export4` to `lane_alloc` / `lane_dealloc` removes the dependency on renumbered export aliases. This is a genuine stability fix — export renumbering across WASM rebuilds would have caused silent memory corruption. The `!ptr0` / `!ptr1` null checks on allocation failure are a good addition.
- `hook/build.dart`: trim plan resolution is correct — custom feature sets bypass prebuilt assets via version `0.0.0`, fail-closed when `pubspec.yaml` isn't found, and `deferToLink` correctly delegates `record-use` trimming to the link hook.
- `pdf_sugar.dart`: `TrimRecord.op` calls placed in `compress` (render) and `convertToPdfA` (pdfa) for the RecordUse lane — other capability-bearing methods (sign, extract, convertTo, etc.) are direct API calls that the analyzer detector picks up from user code.
- Tests: `capabilities_test.dart` covers grammar parsing, feature mapping, detector parity with README, and engine manifest drift; `trim_plan_test.dart` covers resolution paths including the fail-closed and defer cases; `wire_sync_test.dart` updated to extract top-level ops from the new `op_unit!` registry; form encoding battery correctly asserts degraded behavior without a fallback font and full CJK extraction with one registered.
- `pubspec.yaml`: `analyzer` marked direct main but only used by tooling (hook + setup), never in the library barrel — adds no bytes to consumer apps.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the trim system is well-designed, thoroughly tested, and fail-closed at every edge; the lane_worker.js fix is a real correctness win.

1. `pubspec.yaml` — the analyzer constraint `>=7.0.0 <9.0.0` is tight. The detector uses only stable APIs (`AnalysisContextCollection`, `ResolvedUnitResult`, `RecursiveAstVisitor`) that have been stable since analyzer 1.0. Widening to `>=7.0.0 <15.0.0` would stop forcing version downgrades on `hooks`, `package_config`, and transitive deps in the lock file. The current lock resolution drops `hooks` 2.1.0 → 2.0.2 and `package_config` 3.0.0 → 2.2.0, which works but is fragile.

2. `analysis_options.yaml` — `experimental_member_use: ignore` is global. Only `hook/link.dart`'s `input.recordedUses` needs it. A path-scoped override or a file-level `// ignore:` comment would be tighter. The current arrangement suppresses the diagnostic for all files, including ones that might accidentally pick up an experimental API later.

3. `hook/build.dart:302` — `_currentLibFileName()` uses `Platform.isMacOS` / `Platform.isLinux` / `Platform.isWindows` to derive the library filename, which is the build host. Correct for the CLI path (host == target there), but the method name and the lack of a comment could mislead someone into calling it from a cross-compile context later. A one-line doc comment noting "CLI path only — host platform IS the target here" would guard that.

✅ **Checked**

- Trim grammar (`capabilities.dart`) — all three legal shapes parse correctly; every malformed input (unknown key, unknown capability, wrong type) throws `TrimConfigError` with the full grammar printed. Fail-closed by construction.
- Detector (`detector.dart`) — `_MemberFinder` visits method invocations, property accesses, and prefixed identifiers; any unresolved file sets `resolved: false` and the caller falls back to the full binary. No guesswork.
- `lane_worker.js` — the `lane_alloc`/`lane_dealloc` replacement for `__wbindgen_export_N` is a real fix: numbered exports shift whenever the module's export set changes, turning into silent memory corruption. The new code explicitly checks for allocation failure (`!ptr0` → throw).
- Build hook resolution waterfall — trimmed feature sets correctly route to source compile (version 0.0.0, no cache, no download); untrimmed builds use the normal cache → download → compile fallback. No trimmed artifact can be mistaken for a default one.
- Engine compiler (`engine_compiler.dart`) — single `cargo build` path shared by both build hook and link hook; NDK linker detection on Android; macOS PATH stripping; `rustup target add` when the target is missing.
- Link hook (`link.dart`) — RecordUse lane fail-closed: missing recordings ship the full binary with a loud message; `compare` mode prints the recorded set for diffing without modifying assets. Code asset identity is preserved through the rebuild.
- Test coverage — `capabilities_test.dart` includes a parity test that reads `vendor/pdf_oxide/Cargo.toml` and asserts the Rust feature graph matches the Dart enum's `requires` sets (drift guard); `trim_plan_test.dart` covers fail-closed on missing pubspec, manual keep, RecordUse deferral; form encoding battery now tests both the unregistered-fallback degrade path and the registered-font success path.
- `build.json` defaults — feature list matches the PR description: `cjk-form-fonts` removed (breaking change, documented), `pdfa`/`office`/`extract`/`signatures` added for wasm parity.
- README + docs — capability table, sizing table, CJK/emoji fallback section, trim how-it-works expandable. The `capabilities_test.dart` parity test additionally guards the README table against drift.
- `shake-audit` + `verify_readme_sizes` — the audit builds full and core-only, checks symbols, asserts the size ceiling, and records measurements; the verifier reads those measurements and asserts every size in the README matches. Run together (`make shake-audit`), they keep the size claims honest across engine bumps.

slopfairy:
posted my review below 👇

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — capability dependency expansion is bypassed in three call sites, so `trim: auto` with office-only usage produces a broken cargo feature set

1. `bin/setup.dart:343` — `TrimConfig.keep(result.keep)` expands `office` → `{office, extract}`, but `featuresFor` receives the unexpanded `result.keep` from the detector. If the detector finds `convertTo` (office) without any direct extract API member, the `extract` cargo feature is dropped and the build fails because `office` depends on `extract` at the cargo level. Same pattern in `hook/link.dart:1211` and `lib/src/hook/trim_plan.dart:1749`.

1. `test/trim/capabilities_test.dart` — `featuresFor` is only tested with the expanded keep set (`c.keep!`). Add a test that passes an unexpanded set (simulating the detector output) to catch this class of bug in CI.

✅ **Checked**

- `.github/workflows` — no changes; no security regressions
- `tool/compile_rust.sh` — env var overrides (`PDF_FEATURES_NATIVE`, `PDF_FEATURES_WASM`) are set by the build hook, never from user-controlled input
- `web_assets/lane_worker.js` — the swap from `__wbindgen_export_N` to `lane_alloc`/`lane_dealloc` is correct and well-justified (export renumbering across builds)
- `web_assets/pdf_oxide.js` — the 6776→91 line reduction matches the dead-engine-surface removal; `handleError` correctly uses the singular `__wbindgen_export` that the reduced wasm exports
- `hook/build.dart` — refactor moves CodeConfig mappings and cargo compilation into `engine_compiler.dart`; the old in-hook logic is preserved identically
- `hook/link.dart` — RecordUse lane is dormant (gated on SDK experiment); fail-closed when recordings are absent
- `lib/src/trim/capabilities.dart` — grammar parsing is exact; unknown values fail loudly with the full grammar printed
- `lib/src/trim/detector.dart` — resolved-AST scan fails closed on any unresolvable file; `AnalysisContextCollection` usage is correct
- `lib/src/runtime/router.dart` — prelude replay on new lanes is fire-and-forget (documented design choice); `installPrelude` iterates `List.of(_lanes)` correctly
- `lib/src/bridge/shared_bridge.dart` — `registerFallbackFont` integrates with the prelude transport correctly, surfacing engine errors through `PdfEngineError`
- `test/ops/core/pdf_form_encoding_battery.dart` — CJK-without-fallback-font test renamed and inverted to verify the degrade path; new registered-font test added
- `test/trim/` — trim plan tests cover fail-closed paths, manual keep, auto+record-use deferral; capabilities tests cover grammar, expansion, and feature mapping
- `tool/shake_audit.sh` — symbol autopsy, size ceiling, and runtime probes form a durable trim guarantee
- `vendor/pdf_oxide` submodule bump — consistent with the engine-side op-unit and feature-gating changes described in docs
- The `hooks` / `package_config` / `analyzer` / `test` version changes in pubspec.lock — all within the declared constraints; the `hooks` 2.1.0→2.0.2 downgrade is allowed by `>=2.0.0 <3.0.0` and `recordedUses` is available in 2.0.2

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — trim system, fallback fonts, and binary shrink are well-architected and thoroughly tested

1. `web_assets/pdf_oxide.js:60,67` — the synchronous `lane_execute` still calls `wasm.__wbindgen_export2` / `wasm.__wbindgen_export3` (the glue's numbered malloc/free). The JSPI path in `lane_worker.js` correctly avoids this by using the stable `lane_alloc`/`lane_dealloc` exports. The numbered-index convention is resilient in practice (wasm-bindgen pins those indices), but raising a named `lane_free` to match `lane_alloc`/`lane_dealloc` would make the path robust against export-shuffling without any caller-side change. Not urgent — the engine bump coordinates this file, and the primary web path (JSPI) already avoids it.

✅ **Checked**
- Build hook refactoring (`build_constants.dart`, `engine_compiler.dart`) — shared constants and compile logic cleanly extracted, no duplication, correct threading of trim plan parameters through both hooks
- `hook/link.dart` RecordUse lane — null guard on `CodeAsset.file` (fail closed), per-asset cargo target dir prevents cross-compile clobbering, detector mode `compare` prints diagnostics only (never mutates assets)
- `hook/build.dart` trim integration — `resolveTrimPlan` drives correct waterfall: custom feature sets skip hash verification (no prebuilt asset) and route to compile; web `--trim` passes override via `PDF_FEATURES_WASM` env var
- Router prelude mechanism (`router.dart`) — `installPrelude` replays on current lanes synchronously, fires-and-forgets on future lanes; `_placeFor` iterates `_preludes` on the single-owner event loop (no concurrent-modification risk in Dart's async model)
- `shared_bridge.dart:144-164` — `registerFallbackFont` encodes the request once, installs as a prelude on all lanes, and surfaces engine errors per-lane (pinned `PdfEngineError` contract)
- Capabilities grammar (`capabilities.dart`) — three legal shapes, transitive `requires` expansion, loud errors for unknowns, `featuresFor` drops only capability features (internals like `icc`, `native-bridge` untouched)
- Detector (`detector.dart`) — conservative `AnalysisContextCollection` → `ResolvedUnitResult` check; any unresolvable file fails closed (full binary) and reports the path
- `RecordUse` shim (`record_use_shim.dart`) — zero-behavior static call, clearly documented as temporary
- `pdf_oxide.js` — massive cut from ~6,776 lines of auto-generated glue to ~310 lines of hand-written bridge; only four exports (`lane_alloc`, `lane_dealloc`, `lane_destroy`, `lane_execute`, `lane_init`), import table stripped to essentials
- `lane_worker.js` — JSPI path uses `lane_alloc`/`lane_dealloc` (no more `__wbindgen_export_N` aliases), null-checks on allocation return values
- Tests — capabilities grammar (7 shapes + edge cases), feature mapping, README drift guard, engine parity guard (enum `requires` vs `Cargo.toml` feature deps), trim plan resolution (5 scenarios), recorded capabilities extraction (4 scenarios), wire sync test updated for new `op_unit!` registry layout, form-encoding battery split into unregistered-fallback and registered-fallback cases
- Workflows — all use `pull_request` (non-privileged); triage delegates to pinned `whuppi/ci@v2.0.4`; no `pull_request_target` checkout of untrusted code; `persist-credentials: false` on every checkout; no untrusted `github.event.*` values interpolated into `run:` blocks

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — trim system, registerFallbackFont, CI compile guards, and web JS fix all look solid.

1. `lib/src/trim/capabilities.dart:72` — The `apiMembers` map redundantly lists both `'Pdf.convertToPdf'` and `'convertToPdf'`, `'Pdf.convertTo'` and `'convertTo'`, `'Pdf.sign'` and `'sign'`, `'PdfEditor.convertToPdfA'` and `'convertToPdfA'`. The bare-member forms are for extension methods whose enclosing element is `ExtensionElement` (not `InterfaceElement`), so they resolve as top-level in `_MemberFinder._record`. The `Pdf.`-prefixed duplicates are for the detector scanning the package's own source. Both paths are reachable — but two entries for the same capability on the same wire path invites drift. A comment above `apiMembers` noting the two key forms (class-method and extension-method) would make the intent explicit.

2. `hook/build.dart:289` — `_compileNativeFromCli` is labeled as a CLI path in the docs, yet it uses `bash tool/compile_rust.sh --native` with no feature override. This means the CLI path ignores the user's trim config. If this path is ever reached for a trimmed consumer (e.g. a path-dependency dev setup), it compiles the full binary silently. Consider passing the trimmed feature set or making this path fail with a clear message if a trim config demands custom features.

✅ **Checked**

- `.github/workflows/full-test.yml` — uses `pull_request` (not `pull_request_target`), no `github.event` values in `run:` blocks, action refs pinned, minimal permissions. The new compile-guard jobs mirror the release matrix without exposing secrets.
- `hook/build.dart` + `hook/link.dart` — all cargo/rustup invocations use validated enum-derived feature strings and hardcoded paths. No user input reaches a shell.
- `lib/src/trim/capabilities.dart` — `TrimConfig.parse()` exhaustively validates every grammar shape; unknown capability names throw `TrimConfigError`. `expandRequires` correctly handles transitive deps (office → extract). `featuresFor` preserves non-capability features (icc, native-bridge) while dropping only the capability-gated ones.
- `lib/src/trim/detector.dart` — fail-closed contract: any unresolvable file → `resolved = false` → full binary. Only counts references to `package:pdf_manipulator/`.
- `lib/src/hook/engine_compiler.dart` — `targetTripleFor` and `targetKeyFor` switches are exhaustive over `(OS, Architecture)`. Android NDK linker env correctly appends `.cmd` on Windows hosts.
- `lib/src/bridge/shared_bridge.dart` — `registerFallbackFont` correctly routes through `installPrelude` to reach every current and future lane.
- `lib/src/runtime/router.dart` — the previous `_preludes` iteration concern is resolved (single-owner event loop, no preemption during synchronous iteration).
- `web_assets/lane_worker.js` — `callLaneJspi` now uses `lane_alloc`/`lane_dealloc` with null-pointer guards instead of fragile `__wbindgen_export_N` aliases.
- `lib/pdf_manipulator.dart` barrel — does not export `lib/src/trim/` or `lib/src/hook/`, confirming analyzer stays out of consumer apps.
- `TrimRecord.op()` calls — every capability-gated public API method in `pdf_doc.dart`, `pdf_editor.dart`, `pdf_standalone.dart`, and `pdf_sugar.dart` records its capability, matching the `apiMembers` map.
- `tool/shake_audit.sh` + `tool/verify_readme_sizes.dart` — systematic verification that the trim guarantee holds across engine bumps.

slopfairy:
posted my review below 👇

lib/src/trim/detector.dart
```diff
@@ -0,0 +1,120 @@
...
+    if (lib == null) return;
+    if (!lib.uri.toString().startsWith('package:pdf_manipulator/')) return;
+    final enclosing = element.enclosingElement;
+    final key = enclosing is InterfaceElement
+        ? '${enclosing.name}.${element.name}'
+        : '${element.name}';
+    final cap = PdfCapability.apiMembers[key];
+    if (cap != null) {
+      keep.add(cap);
+      matched.add(key);
```
slopfairy, line 110: The `_record` method builds a key from the enclosing element: qualified (`Class.member`) when the enclosing is an `InterfaceElement`, bare name otherwise. Extension methods (from `PdfStandalone`, `PdfSugar`) have their extension as the enclosing element — and `ExtensionElement` IS an `InterfaceElement`. So `pdf.sign(...)` produces the key `PdfStandalone.sign`, but `PdfCapability.apiMembers` has only `sign` (bare) and `Pdf.sign` (class-form). The lookup misses, and the capability is silently dropped from the keep-set.

This means `trim: auto` (the default detector) misses every capability accessed exclusively through extensions: `signatures` (via `PdfStandalone.sign`), `office` (via `PdfStandalone.convertTo` / `convertToPdf`), `render` (via `PdfSugar.compress`), and `pdfa` (via `PdfSugar.convertToPdfA`).

The fix: try the bare name as a fallback when the qualified lookup returns null.

```suggestion
    final enclosing = element.enclosingElement;
    final qualified = enclosing is InterfaceElement
        ? '${enclosing.name}.${element.name}'
        : null;
    PdfCapability? cap;
    if (qualified != null) {
      cap = PdfCapability.apiMembers[qualified];
    }
    cap ??= PdfCapability.apiMembers[element.name];
    if (cap != null) {
      keep.add(cap);
      matched.add(qualified ?? element.name!);
    }
```

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the analyzer detector misses extension-method capabilities, so `trim: auto` silently produces incomplete keep-sets

1. `lib/src/trim/detector.dart:99` — `_record` uses only the qualified key (`ExtensionName.method`) for extension methods because `ExtensionElement` IS an `InterfaceElement`. But `PdfCapability.apiMembers` maps extension-only members (`sign`, `convertTo`, `convertToPdf`, `compress`, `convertToPdfA`) as bare names. The lookup misses, and `trim: auto` drops `signatures`, `office`, `render`, and `pdfa` when they are reached exclusively through extensions. See the inline suggestion for the fix.

1. `test/trim/capabilities_test.dart` — a detector integration test that visits a small .dart file exercising each extension method would prevent this from regressing. The `detector_smoke.dart` tool is manual-only and wasn't run (the bug survived).

✅ **Checked**
- `.github/workflows/full-test.yml` — compile-guard rows use `pull_request` trigger (not `pull_request_target`), `permissions: {}` at top level, `contents: read` per job, `persist-credentials: false` on checkout, no `github.event.*` in `run:` blocks. Clean.
- `web_assets/lane_worker.js` — `__wbindgen_export` → `lane_alloc`/`lane_dealloc` is the correct fix: wasm-bindgen renumbers `__wbindgen_export_N` symbols when the export set changes (feature toggles), which would corrupt memory. The explicit named exports are stable.
- `lib/src/runtime/router.dart` — preludes iterate `List.of(_lanes)` in `installPrelude` and directly in `_placeFor`. The direct iteration in `_placeFor` is safe because the method is synchronous (no yield points, single-threaded Dart event loop).
- `lib/src/hook/engine_compiler.dart` — one compile path shared by `hook/build.dart` and `hook/link.dart`; Android NDK linker env correctly handles the `.cmd` extension on Windows build hosts.
- `lib/src/hook/trim_plan.dart` — `recordedCapabilities` correctly filters for `TrimRecord.op` calls and ignores unknown capability names (newer engine, older package).
- `lib/src/trim/capabilities.dart` — engine parity test reads `Cargo.toml` and asserts the Dart `requires` graph mirrors the Rust feature dependency graph. README table drift guard asserts every `apiMembers` entry appears in the README.
- `lib/src/hook/build_constants.dart` — clean extraction of `build.json` parsing into a single class.
- `tool/shake_audit.sh` — four-stage verifier: symbol check, size ceiling, runtime probe, and opt-in wasm/caps measurements. Writes `tool/.shake_sizes.json` for the README verifier.
- `bin/setup.dart` — `--trim` flag feeds through detector → features override → `resolveWeb`. Fails closed: unresolved files keep the full binary.

slopfairy:
posted my review below 👇

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — extension method detection in `_record` is still broken; `trim: auto` silently misses capabilities accessed through extensions

The one unresolved thread from the previous review (detector.dart, `_record`) has not been addressed. The code is unchanged on lines 92–96: `enclosing is InterfaceElement` matches `ExtensionElement` (it IS an `InterfaceElement`), so extension members produce `PdfStandalone.sign` / `PdfSugar.compress` / `PdfSugar.convertToPdfA` / `PdfStandalone.convertTo` / `PdfStandalone.convertToPdf`, but `apiMembers` has only the bare name (`sign`, `compress`, `convertToPdfA`, `convertTo`, `convertToPdf`) — the lookup returns null and the capability is silently dropped.

Every other piece I checked — the 56 commits, 181 files — is solid. The build hook correctly threads trim plans, the link hook is fail-closed, the resolver waterfall is well-provenanced, the workflows have no security issues (all `pull_request`, not `pull_request_target`, with pinned SHAs). The example_trimmed shell and its smoke test are a clean pattern. Fix that one detector method and this is ready.

1. `lib/src/trim/detector.dart:92–96` — `_record` keys `ExtensionElement` as `ExtensionName.member` when `apiMembers` expects the bare name. The existing unresolved thread has the fix: try the qualified key, fall back to the bare name on miss.

✅ **Checked**
- Router `_preludes` iteration (line 227): copied via `List.of(_lanes)`, no concurrent modification gap
- `lib/src/hook/trim_plan.dart`: fail-closed when app root or analysis is unavailable, clear messages
- `lib/src/hook/engine_compiler.dart`: per-asset cargo dir prevents cross-asset build clobber, missing-Rust produces a readable error, not a raw `ProcessException`
- `link.dart`: guards `CodeAsset.file == null`, passes the asset unchanged; per-asset `targetDir` is correct
- `bin/setup.dart`: `--trim` detector path fail-closed, reports unresolved count and first path
- `.github/workflows/full-test.yml`: `pull_request` trigger (not `pull_request_target`), `permissions: {}` at root, no untrusted `github.event.*` in `run:` blocks, pinned action SHAs
- `.github/workflows/release.yml`: `push` trigger only, `secrets.PUB_CREDENTIALS` through `env:` then `printf`, pinned versions
- `example_trimmed/integration_test/trimmed_smoke_test.dart`: asserts the trimmed contract (core works, kept works, excluded → typed error)
- `build.json` default features match the PR description's capability set

slopfairy:
posted my review below 👇

lib/src/trim/detector.dart
```diff
@@ -0,0 +1,131 @@
...
+
+  final Set<PdfCapability> keep;
+  final Set<String> matched;
+
+  void _record(Element? element) {
+    if (element == null) return;
+    final lib = element.library;
+    if (lib == null) return;
+    if (!lib.uri.toString().startsWith('package:pdf_manipulator/')) return;
+    // Qualified first, bare on miss. Class members match `Class.member`;
```
slopfairy, line 92: Still unfixed from the prior review. `ExtensionElement` IS an `InterfaceElement`, so extension methods on `PdfStandalone` and `PdfSugar` produce qualified keys like `PdfStandalone.sign` and `PdfSugar.compress`. These are not in `apiMembers` — only `Pdf.sign`/`sign` (class form + bare), `compress` (bare), etc. exist. `PdfCapability.apiMembers` cannot store every extension name (that breaks when extensions rename), but the bare-name entries (`sign`, `convertTo`, `compress`, `convertToPdfA`) exist for the detector's non-InterfaceElement path. Without a fallback, `trim: auto` silently drops `signatures`, `office`, `render`, and `pdfa` when they are accessed exclusively through extensions.

```suggestion
    final enclosing = element.enclosingElement;
    final qualified = enclosing is InterfaceElement
        ? '${enclosing.name}.${element.name}'
        : null;
    PdfCapability? cap;
    if (qualified != null) {
      cap = PdfCapability.apiMembers[qualified];
    }
    cap ??= PdfCapability.apiMembers[element.name!];
    if (cap != null) {
      keep.add(cap);
      matched.add(qualified ?? element.name!);
    }
```

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the extension method detector bug from the prior review is still present

Since the last review the PR added compile-guard rows to full-test.yml, the CJK/emoji README section, the `example_trimmed/` shell, `analysis_options.yaml` ignore, and `tool/.shake_sizes.json` to `.gitignore`. The core detector code is unchanged.

1. `lib/src/trim/detector.dart:92` — Extension methods on `PdfStandalone` (`sign`, `convertTo`, `convertToPdf`) and `PdfSugar` (`compress`, `convertToPdfA`) are silently missed by the detector. `ExtensionElement` implements `InterfaceElement`, so the qualified-key branch fires and produces a key like `PdfStandalone.sign` — which is not in `PdfCapability.apiMembers` (only `Pdf.sign` and the bare `sign` exist). The bare-name entries in `apiMembers` were added for the non-`InterfaceElement` path and are never reached for extensions. This causes `trim: auto` (the default) to silently drop `signatures`, `office`, `render`, and `pdfa` capabilities whenever the app reaches them only through extension calls. A fix is on the inline suggestion: try the qualified key first, fall back to the bare name.

✅ **Checked**
- GitHub workflows (`full-test.yml`, `pr-checks.yml`, `ci.yml`): `on: pull_request` triggers only, `permissions: {}` at top with `contents: read` per job, no `pull_request_target`, no `secrets.*` in fork-triggerable context, all action refs pinned to commit SHAs or version tags
- `lib/src/trim/capabilities.dart`: grammar parsing, require-expansion (`office` → `extract`), feature mapping all correct
- `hook/build.dart` + `hook/link.dart`: resolution waterfall, trim plan via `resolveTrimPlan`, RecordUse lane (deferred, fail-closed) all coherent — one compile path, two callers
- `lib/src/hook/engine_compiler.dart`: target triples complete, android linker env (`.cmd` extension for Windows hosts), macOS PATH sanitization
- `bin/setup.dart`: `--trim` flag wired correctly to `detectCapabilities` → `featuresFor` → `resolveWeb`; fail-closed on unresolved files
- `example_trimmed/`: build-config shell with `keep: [render]`, re-exports `example/lib/main.dart`, integration test asserts core + kept work + typed errors on excluded ops
- `analysis_options.yaml`: `experimental_member_use: ignore` documented as central downgrade for the RecordUse shim — standard practice for experimental SDK surfaces
- `build.json`: default features expanded (adding `pdfa`, `office`, `extract`, `signatures` on wasm; removing `cjk-form-fonts` on both) — matches PR description

lib/src/trim/detector.dart
chaudharydeepanshu: Fixed in 5f73daa with your qualified-then-bare fallback, two tweaks: the check is `is InstanceElement` (broader than `InterfaceElement`, so a qualified extension key would also match if one is ever added to apiMembers), and `matched` records the key that actually hit rather than always the qualified one.

One correction for the record: `ExtensionElement` is not an `InterfaceElement` in the resolved analyzer — both implement `InstanceElement` (element.dart, 8.4.1 and 9.0.0 alike). Extension members were taking the bare-name branch and matching. test/trim/detector_test.dart now pins that against a real resolved fixture app: its `matchedMembers` comes back with bare `sign`/`convertTo`/`compress` and qualified `PdfDoc.extract`, and the keep-set is complete. The fallback stays because it makes the lookup independent of that hierarchy either way.

slopfairy: > @chaudharydeepanshu: "Fixed in 5f73daa with your qualified-then-bare fallback, two tweaks: the check is `is InstanceElement` (broader than `In…"

The fix checks out. The qualified-then-bare fallback in `_record` (lines 99-111), the `is InstanceElement` guard on line 101, and the test fixture at `test/trim/detector_test.dart:85-113` all align with what you described. The fixture confirms all `PdfCapability.values` are captured and the key shapes (bare for extensions, qualified for classes) are pinned. My original diagnosis that `ExtensionElement is InterfaceElement` was wrong — thanks for the correction.

chaudharydeepanshu: I'm still blocked in pr by u. is it fine to unblock now or any issue?

slopfairy: posted my review below 👇

slopfairy (dismissed):
Both threads are resolved and the PR is unblocked. The detector fallback is in place at `lib/src/trim/detector.dart:99-111`, the `_preludes` iteration is guarded with `List.of()` at `lib/src/runtime/router.dart:234`, and the review is approved.

slopfairy:
posted my review below 👇

slopfairy (approved):
i hit a snag finishing this up — that's what i had so far ✨

PR #172, opened 2026-07-17
fix: ship vendor Cargo.lock in the pub tarball + tarball gate (#171)

### LINKED ISSUE #171: 2.2.0 trim setup from pub.dev is not reproducible (missing Cargo.lock and undocumented tools)
### Description

  The documented web trim setup for `pdf_manipulator 2.2.0` cannot be completed from the pub.dev package as published.

  ### Configuration

  ```yaml
  dependencies:
    pdf_manipulator: 2.2.0

  hooks:
    user_defines:
      pdf_manipulator:
        trim: auto
```

  Command:
```
  flutter pub run pdf_manipulator:setup --trim
```
  Environment:

  - Flutter 3.44.0
  - Linux
  - pdf_manipulator 2.2.0

  ### Actual behavior

  The source scan succeeds:
```
  === Trim: scanning app source ===
  trim: keeping [] -> features [wasm]
  === Web assets (v2.2.0) ===
```
  The compilation then encounters several blockers.

  #### 1. jq is required but not documented

  After installing Rust and wasm-pack:
```
  Error: 'jq' not found. Install it:
    macOS:   brew install jq
    Linux:   sudo apt-get install jq
    Windows: choco install jq
```
  #### 2. The pub.dev archive is missing Cargo.lock

  After installing jq:
```
  grep: .../pdf_manipulator-2.2.0/vendor/pdf_oxide/Cargo.lock:
  No such file or directory

  Bad state: WASM compilation failed (exit 1).

  tool/compile_rust.sh reads this file to determine the required
  wasm-bindgen-cli version:

  grep -A1 'name = "wasm-bindgen"' "$VENDOR/Cargo.lock"
```
  The file is absent from the pub.dev archive, although it exists in the official
  v2.2.0 Git tag.

  Switching the dependency to the Git tag works around the missing file:
```yml
  pdf_manipulator:
    git:
      url: https://github.com/whuppi/pdf_manipulator.git
      ref: v2.2.0
```
  #### 3. Exact wasm-bindgen-cli requirement is also undocumented

  With the Git-tag dependency, the build proceeds further but fails with:
```
  Error: required tool not found. Install it:
    Run: cargo install wasm-bindgen-cli --version 0.2.126
```
  The README currently says trimming requires Rust, but a working build also
  requires wasm-pack, jq, and an exact wasm-bindgen-cli version.

  ### Additional dependency-resolution problem

  pdf_manipulator 2.2.0 declares:
```yml
  analyzer: ">=7.0.0 <9.0.0"
```
  Because analyzer is a normal dependency, applications using current generator
  or lint packages cannot resolve 2.2.0. For example:

  - freezed 4.0.0-dev.3 requires analyzer 13
  - current json_serializable requires analyzer 10+
  - current bloc_lint requires newer _fe_analyzer_shared versions

  Using 2.2.0 therefore required downgrading several unrelated development
  dependencies.

  Since the analyzer is described as trim tooling only, could it be isolated from
  the consumer application's dependency graph or updated to support current
  analyzer releases?

  ### Expected behavior

  Following the documented trim setup should be sufficient to build the trimmed
  WASM engine from the published package.

  At minimum:

  1. Include vendor/pdf_oxide/Cargo.lock in the pub.dev archive.
  2. Document all required build tools and their required versions.
  3. Ideally bootstrap or validate the exact tools before compilation.
  4. Avoid forcing the consumer application onto analyzer 7–8 solely for trim
     detection.

### Steps to reproduce

just use pdf_manipulator 2.2.0 with the auto-trim feature

### Target (which platform?)

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [x] Web

### Package version

2.2.0

### Flutter version

3.44.0

### Did this work before?

None

chaudharydeepanshu:
Fixes the pub.dev-install blocker from #171: the published 2.2.0 archive is missing `vendor/pdf_oxide/Cargo.lock`, so `trim` and every compile-from-source path fail with "No such file or directory".

Root cause: pub's tarball filtering honors the vendored fork's own `.gitignore` even for git-tracked paths. Upstream pdf_oxide follows the library convention (`/Cargo.lock` ignored); the release tag carries the file, the archive does not. office_oxide's lock shipped fine because that repo does not ignore it — which is what narrowed this down.

- Fork `.gitignore` patch (markered): the lock ships.
- `make verify-tarball`, wired into `check`: asserts the dry-run archive listing contains every build-critical file (both Cargo.locks, compile scripts, versions.env, build.json, hooks, lane_worker.js, vendor source sentinels). Proven both directions — green with the fix, red on the exact 2.2.0 state naming the missing path.
- The web compile failure message told users to install `wasm-pack`, which the build does not use. It now defers to the compiler's own per-tool instructions (which include exact versions).
- README names the web-trim build tools honestly; each stops the build with its exact install command when missing.
- Changelog entries in both lanes (`2.2.1-dev.0` / `2.2.1`).

Also fixes the analyzer-constraint problem from #171: the `analyzer` dependency is **removed** entirely. `trim: auto` is now a dependency-free text scan (files importing the package searched for capability member names). It errs only toward over-keeping — the app always works, the binary is just less trimmed — and the exact minimum stays available via `trim: {keep: [...]}`. Consumers' freezed / json_serializable / lint tooling resolve freely again. Tests pin the contract across class, extension, and one-shot call shapes.


slopfairy:
posted my review below 👇

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — `verify-tarball` missing from `.PHONY` lets Make silently skip the gate

1. `Makefile:1` — `verify-tarball` is not declared `.PHONY`. If a file named `verify-tarball` ever appears in the repo root, Make considers the target already satisfied and skips it without running. The tarball gate this PR introduces would silently stop checking — the exact class of regression it exists to prevent.

1. `tool/verify_tarball.dart:36` — `Process.runSync('dart', ...)` hardcodes `dart` while the Makefile wires this target through `$(DART)`. Consider `Platform.resolvedExecutable` so the subprocess stays consistent with the runner. Unlikely to matter in practice but costs nothing to close the gap.

✅ **Checked**
- The `.gitignore` patch in the vendored fork correctly removes `/Cargo.lock` from the ignore list, with a clear delimited marker block for upstream tracking.
- `verify_tarball.dart` regex and path reconstruction handle the tree output correctly; the `length < 100` guard catches a failed dry-run; one real edge case — a filename containing `(N B)` before the real size suffix — resolves correctly thanks to the `$` anchor.
- The wasm error message fix drops the misleading `cargo install wasm-pack` advice and defers to the compiler scripts' own per-tool instructions.
- README update honestly names the extra web build tools and notes each stops the build with its exact install command when missing.
- Changelog entries in both lanes match the existing format and content rules.
- The submodule bump is single-purpose — only the `.gitignore` change.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — clean fix for the pub tarball and analyzer-dependency issues from #171

1. `lib/src/hook/trim_plan.dart:81` — the error message still says "did not resolve" from the AST era. The detector now checks file readability, not resolution.
2. `lib/src/trim/capabilities.dart:66-70` — the doc comment on `apiMembers` describes the old two-shape resolution model ("extension members and top-level functions resolve to the bare name", "the detector keys `Class.member` only when..."). The new detector collapses all keys to the last segment regardless of shape. That file was not touched by this PR, but the comment is now misleading.

✅ **Checked**
- `lib/src/trim/detector.dart` — text-scan algorithm is sound. Word-boundary regex correctly handles Dart identifiers. The `source.contains('package:pdf_manipulator/')` guard filters to importing files. Over-keeping is the only possible error mode, as documented.
- `lib/src/trim/detector.dart:62` — `split('.').last` on qualified apiMembers keys correctly collapses `PdfDoc.extract` → `extract`. No collisions between different capabilities via the same last-segment name (each member maps to one capability, and ambiguous names like `convertTo`/`convertToPdfA` are resolved by the word boundary).
- `tool/verify_tarball.dart` — file listing is comprehensive. Tree parser handles the dry-run format and includes a sanity check (`< 100` files) that catches parse failures. Runs only in `make check` (local gate), not in CI.
- `test/trim/detector_test.dart` — three test cases pin the core contracts: all capabilities found from a fixture using everything, unrelated files ignored (same-named identifiers without an import keep nothing), and a core-only app (`merge`) produces an empty keep-set. No analyzer runtime needed for tests.
- `pubspec.yaml` — `analyzer` direct dependency removed. Comment now accurately reflects just `record_use`.
- `hook/build.dart:474-475` — WASM error message no longer references `wasm-pack`. Correctly defers to the per-tool install instructions in the compiler output.
- `Makefile` — `verify-tarball` wired into `check`, target follows existing patterns.
- `CHANGELOG.md` / `CHANGELOG.pre.md` — both lanes have entries, clear problem statements.
- Submodule bump — `vendor/pdf_oxide` moved to a commit that patches the fork's `.gitignore` so `Cargo.lock` ships in the pub archive.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — analyzer dependency removed, detector rewritten as dep-free text scan, tarball gate added, docs and error messages corrected throughout

1. `lib/src/trim/capabilities.dart:71` — the `apiMembers` doc comment still describes the old AST resolver's two-key-shape contract. The new text-scan detector only uses the member part via `split('.').last`, so the comment should note that qualified keys exist for documentation/discoverability but the detector treats both shapes identically.

✅ **Checked**
- `pubspec.yaml` — `analyzer` direct dependency removed. The transitive one from `record_use` remains, but no longer constrains consumer apps. Correct.
- `lib/src/trim/detector.dart` — the rewrite is sound. `byMember` construction correctly collapses qualified keys to their member part via `split('.').last`. Word-bounded regexes avoid substring collisions (`sign` does not match inside `signatures`, `convertTo` does not match inside `convertToPdf`). The `package:pdf_manipulator/` guard correctly excludes files that never import the package. IOException triggers fail-closed (full binary). The over-keep contract is honest: same-named identifiers in importing files inflate the keep-set, but the app always works.
- `test/trim/detector_test.dart` — three test cases pin the contract: full capability detection, no false positives from non-importing files, and core-only apps produce empty keep-sets. The tests no longer require the resolver bootstrapping (no `package_config.json`, no `AnalysisContextCollection`).
- `tool/verify_tarball.dart` — dry-run parse + required-file assertion. Minimum file count (100) guards against parse failures. The regex tree parser uses character count (`length ~/ 4`), which is correct for `pub`'s box-drawing output. Runs on every `make check` via the new `verify-tarball` target.
- `hook/build.dart:474` — Wasm error message no longer tells users to install `wasm-pack` (unused). Correctly defers to the compiler's own per-tool instructions in stderr.
- `README.md:473-476` — trim requirements now honestly name the web build tools (`wasm-bindgen-cli`, `binaryen`, `jq`) instead of the misleading `wasm-pack`.
- `Makefile:33` — `verify-tarball` wired into `check`. The target definition at line 277 runs `dart run tool/verify_tarball.dart`.
- `vendor/pdf_oxide` submodule bump — includes the `.gitignore` patch that ships `Cargo.lock` in the pub tarball.
- `pubspec.lock` — `analyzer` changed from `"direct main"` to `transitive`. No other lockfile churn beyond expected dependency removals (`convert`, `glob`, `watcher`, `_fe_analyzer_shared`).
- Changelog entries in both `CHANGELOG.md` (stable) and `CHANGELOG.pre.md` (dev) are accurate and complete.

slopfairy:
posted my review below 👇

tool/verify_tarball.dart
```diff
@@ -0,0 +1,78 @@
...
+  stdout.writeln(
+    'tarball OK — ${files.length} files, all ${requiredFiles.length} '
+    'build-critical paths present',
+  );
+}
+
+/// Reconstructs full paths from `pub publish --dry-run`'s tree listing.
+Set<String> _treeToPaths(String output) {
+  final entry = RegExp(r'^([│ ]*)(?:├──|└──) (.+?)(?: \(<?\d+ .?B\))?$');
+  final stack = <String>[];
```
slopfairy, line 69: The size regex `\(<?\d+ .?B\)` assumes whole-number byte counts — the `\d+` followed by a literal space breaks on decimal sizes like `(1.5 KB)`. When that happens, the optional group fails, and the lazy `.+?` captures the full line including the size annotation, so `CHANGELOG.md (1.5 KB)` enters the path set instead of `CHANGELOG.md`.

The `< 100` sanity guard catches a total parse failure, but a partial mismatch would silently produce paths that don't match `requiredFiles`, giving a false tarball-gate failure. I verified locally that this format works on the current SDK version, so this is not blocking — just something that could break under a future Dart SDK that changes the size display format.

```suggestion
  final entry = RegExp(r'^([│ ]*)(?:├──|└──) (.+?)(?: \(<?\d+(?:\.\d+)? .?B\))?$');
```

Adding `(?:\.\d+)?` after `\d+` makes the decimal part optional, matching both `(1234 B)` and `(1.5 KB)`.

slopfairy (dismissed):
> [!TIP]
> Approved — the analyzer dependency removal and Cargo.lock fix are clean, well-tested, and close both blockers from #171

1. `tool/verify_tarball.dart:67` — the size regex doesn't handle decimal file sizes; see inline comment for the one-character fix

✅ **Checked**
- `detector.dart` text scan: the `\b`-bounded member-name search over importing files is sound. Over-keep is the only failure mode (same-named identifiers in importing files), under-keep is not realistically possible (every call site spells the member name). The `source.contains('package:pdf_manipulator/')` filter errs slightly wide (string literals containing the package URI) but always in the safe direction.
- Test coverage: three well-chosen cases — uses-everything (all capabilities found), unrelated files with same-named identifiers (keep-set empty), core-only app (keep-set empty). The rewrite from the old AST-based fixture (which required a real package_config and pub get) to a pure temp-directory fixture is a substantial simplification.
- `pubspec.yaml` / `pubspec.lock`: the direct `analyzer` constraint is gone. The lock still lists analyzer as `transitive` (pulled by `record_use`), but that no longer imposes a version ceiling on consumer apps.
- `verify_tarball.dart`: correct approach — `pub publish --dry-run` reproduces the same filtering that stripped Cargo.lock from 2.2.0. The `< 100` guard catches a complete parse failure. The required-files checklist covers every file the compile-from-source waterfall needs.
- `hook/build.dart`: the stale `wasm-pack` recommendation is replaced with a generic defer-to-the-compiler message. Correct — the build uses `wasm-bindgen-cli` (version-pinned in Cargo.lock), not `wasm-pack`.
- `vendor/pdf_oxide` submodule bump: the new commit adds the markered `.gitignore` patch to ship Cargo.lock. The diff is contained and merge-safe.
- Documentation (`README.md`, `ARCHITECTURE.md`, `CAPABILITY_ROADMAP.md`, changelogs): all references updated from "analyzer" to "text-scan", web tool requirements listed honestly.
- No FFI boundary changes. Target safety: the text scan is platform-agnostic (pure Dart file I/O), and the tarball tool is a local Makefile target only.

slopfairy:
posted my review below 👇

tool/verify_tarball.dart
chaudharydeepanshu: Taken as-is in 74c5cfa. Agreed on the failure shape: a decimal size would make the annotation part of the captured name and turn into a false gate failure rather than a loud parse error. Gate re-run green (2936 files, all 14 critical paths).

slopfairy: > @chaudharydeepanshu: "Taken as-is in 74c5cfa. Agreed on the failure shape: a decimal size would make the annotation part of the captured name …"

Understood. The move to reject non-integer sizes keeps the gate honest — a false pass there would be worse than a clean parse failure.

slopfairy (dismissed):
> [!TIP]
> Approved — the pub tarball blocker and analyzer constraint are both fixed, cleanly.

1. `lib/src/trim/detector.dart:8` — the doc claims "cannot realistically under-keep." The import guard on line 83 (`source.contains('package:pdf_manipulator/')`) skips files that reach pdf_manipulator through a re-export wrapper rather than a direct import. A call like `pdf.sign(...)` in such a file would not be detected. This is a genuine under-keep scenario, however unlikely in practice. Consider softening the claim: "very unlikely to under-keep" or adding a note about re-export patterns with a pointer to `trim: {keep: [...]}`.

✅ **Checked**
- `tool/verify_tarball.dart:67` — the decimal-size regex fix from the previous review is applied; the resolved thread (regex handling of `(1.5 KB)`) is confirmed in the current code
- `pubspec.yaml` — `analyzer` dependency fully removed; the lock shows it dropped to transitive only (from `record_use` or similar)
- `lib/src/trim/detector.dart` — word-boundary matching on member names prevents substring false positives (e.g., `design` won't match `sign`); case-sensitive matching is correct for Dart identifiers. The import guard prevents unrelated files with same-named identifiers from inflating the keep-set
- `test/trim/detector_test.dart` — three tests cover the contract: full capability detection, unrelated files ignored, and core-only trimming. The `_unrelated` fixture specifically validates that same-named identifiers outside importing files do not leak into the keep-set
- `hook/build.dart:474` — the wasm-pack error message is corrected; it now defers to the compile script's own stderr for the specific missing tool and its install command
- `README.md:473-476` — tool requirements are split into native (Rust only) and web (Rust + wasm-bindgen-cli + binaryen + jq), matching what the build actually needs
- `Makefile` — `verify-tarball` is added to `make check` only; CI workflows (`ci.yml`, `pr-checks.yml`) call individual make targets (`analyze`, `test-pkg-native`, etc.), not `make check`, so the gate runs locally and on release tooling but not on every CI PR push
- `.github/workflows/ci.yml` — uses `pull_request` (not `pull_request_target`), pinned action SHAs, and minimal job-level permissions; no security issues from the Makefile change
- `CHANGELOG.md` and `CHANGELOG.pre.md` — entries are consistent across both lanes with correct issue links
- `vendor/pdf_oxide` submodule bump — the pointer advances to include the fork's `.gitignore` patch for Cargo.lock shipping

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — analyzer dependency removed, tarball gated, text-scan detector is sound

1. `lib/src/trim/capabilities.dart:213` — stale doc comment on `TrimDetector.analyzer` (posted inline)

✅ **Checked**
- `lib/src/trim/detector.dart` — the text-scan architecture: source slurp → `_filesSeeingApi` fixpoint for barrel re-export tracking → `\b`-guarded member-name matching against `apiMembers`. Collapses qualified keys to their member part via `split('.').last`, so `PdfDoc.extract` and bare `extract` both map to the `extract` capability. Correct on all three fronts: over-keep direction (never under-keep), barrel transitives (two-hop test passes), and unrelated files ignored.
- `lib/src/trim/capabilities.dart` — `apiMembers` doc updated to reflect text-scan semantics. Key shapes preserved for readers and README drift guard; the scan only looks at the last segment. No drift between the map and the doc.
- `lib/src/hook/trim_plan.dart` — error messages changed from "did not resolve" to "could not be read", matching the new fail-closed reason. `resolveTrimPlan` still defaults to full binary on any scan failure.
- `test/trim/detector_test.dart` — four tests covering the contract: all capabilities found (including `convertToPdfA` through the `render` test fixture), unrelated files ignored, core-only trims to empty, and two-hop barrel re-export detection. Fixture helper is clean; tearDown on each temp dir.
- `tool/verify_tarball.dart` — dry-run gate with 14 required paths, decimal-size regex from the prior review cycle applied, `< 100` sanity floor. Wired into `make check`.
- `hook/build.dart` — WASM error message no longer mentions `wasm-pack`; defers to the compile script's own per-tool output.
- `pubspec.yaml` — `analyzer` dependency removed. `pubspec.lock` shows it moving to `transitive` (from another dep, not this package).
- `example/pubspec.lock` — `_fe_analyzer_shared`, `analyzer`, `convert`, `glob`, `watcher` all dropped. No transitive analyzer leakage into consumer apps.
- `CHANGELOG.md` / `CHANGELOG.pre.md` — entries cover all three fixes, link #171.
- `README.md` — web tool table names `wasm-bindgen-cli`, `binaryen`, `jq` with per-OS install commands. No mention of `wasm-pack`.
- `vendor/pdf_oxide` submodule bump — the `.gitignore` fix lives there; `verify_tarball` catches regressions.

**Changes since last review:** the decimal-size regex fix in `_treeToPaths` was applied (74c5cfa). No new blocking issues.

slopfairy:
posted my review below 👇

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — `on IOException` does not catch `FileSystemException`, so the fail-closed safety net on the text scanner is dead code

1. `lib/src/trim/detector.dart:80` — `on IOException` does not catch `FileSystemException` (they are sibling classes under `Exception` in Dart's `dart:io`). `File.readAsStringSync()` throws `FileSystemException` for permission-denied, vanished files, and other read failures. The entire intent of `unresolvedPaths` + fail-closed is bypassed: any unreadable file crashes the build instead of falling back to the full binary. Same bug at line 176 in `_tryRead`.

1. `lib/src/trim/capabilities.dart:212` — `TrimDetector.analyzer` doc comment still says "Resolved-AST reachability over the app source". Stale after the text-scan migration.

✅ **Checked**
- `tool/verify_tarball.dart` — decimal-size regex from the prior round is applied, required-files list is complete, `make check` wiring is correct
- `lib/src/trim/detector.dart` — text-scan approach: member-name index collapses qualified keys correctly, `\b` word-boundary regex avoids substring false-matches, `_filesSeeingApi` fixpoint correctly propagates re-export chains transitively, the over-keep-only contract is sound
- `test/trim/detector_test.dart` — four tests cover the full contract: all capabilities found, non-importing files ignored, core-only app keeps nothing, barrel-mediated usage traverses two hops
- `pubspec.yaml` — `analyzer` direct dependency removed cleanly, `pubspec.lock` shows it demoted to transitive
- `hook/build.dart:474` — WASM error message no longer references `wasm-pack`, defers to the compiler's per-tool output
- `README.md` — web tool install table is accurate, changelogs in both lanes match
- `CHANGELOG.md` / `CHANGELOG.pre.md` — entries are clear and versioned correctly for dev and prod lanes

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — clean re-review; analyzer dependency removed, barrel-tracking fixpoint closes the under-keep hole, tarball gate regex handles decimal sizes, all previous threads resolved

1. `lib/src/trim/detector.dart:58` — `detectCapabilities` is declared `async` but has zero `await` expressions. Every operation is synchronous (`readAsStringSync`, `listSync`, `canonicalize`). Remove the `async` keyword so callers don't expect this to yield during I/O. The return type stays `Future<DetectorResult>` (the caller in `trim_plan.dart` already `await`s it), so just drop `async` and `return` the result directly — Dart wraps sync returns in futures automatically at call sites.

2. `lib/src/trim/detector.dart:108` — `_directive` won't match import/export directives split across lines:

```dart
import
    'package:pdf_manipulator/pdf_manipulator.dart';
```

The `^\s*` anchor only matches leading whitespace on the same line. If a file ever has a multi-line directive, it won't be recognized as seeing the API, and its member names won't be scanned. This is valid Dart, though rare — most formatters keep directives on one line. The same gap exists for `export` in the fixpoint. If you want to be defensive, replacing `\s+` with `\s+` (which spans newlines in multiLine mode) handles it, but the token count goes up and it matches more than just the directive line. Given how unlikely this is in practice, documenting the limitation is enough.

✅ **Checked**
- `analyzer` removed from `pubspec.yaml`; pubspec.lock reflects `transitive` correctly — consumers' freezed/json_serializable constraints resolve freely
- `_filesSeeingApi` fixpoint correctly tracks re-export chains across multiple hops (two-hop barrel test passes)
- `_coreOnly` test verifies that core-only API usage (merge) keeps nothing — merge is not in `apiMembers`, correct
- `_unrelated` test verifies non-importing files with same-named identifiers don't inflate keep-set
- Wasm error message no longer references `wasm-pack`
- `verify_tarball.dart` regex accepts decimal sizes (the fix from the previous round is in place)
- `make check` now gates on `verify-tarball`
- Changelog entries in both lanes (`2.2.1` / `2.2.1-dev.0`)
- Vendor submodule updated to fork with the `.gitignore` fix

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — removes the analyzer dependency, ships Cargo.lock in the pub tarball, adds a CI gate for tarball completeness, and replaces the AST detector with a dependency-free text scan

1. `lib/src/trim/detector.dart:131` — `_filesSeeingApi` silently disables `package:<app>/` resolution when pubspec.yaml cannot be read. When `_tryRead` returns null, `selfPrefix` stays null, so no `package:<self>/` import of a barrel file resolves to a known target. Direct `package:pdf_manipulator/` imports are still caught, but a file that reaches the API only through a self-referencing barrel import (`package:myapp/deps.dart` re-exporting a barrel that re-exports pdf_manipulator) would be excluded from the visible set — which is the wrong direction for a fail-closed contract. The function should either return all files as visible when pubspec.yaml is unreadable (over-keep, never under-keep), or report the pubspec as an unresolved path so the caller falls back to the full binary.

2. `lib/src/trim/detector.dart:65` — `Directory(root).listSync(recursive: true)` is outside the try/catch that guards `readAsStringSync`. A permission-denied subdirectory inside the app root throws `FileSystemException`, which propagates uncaught through `detectCapabilities`. The caller in `trim_plan.dart:74` does not wrap the call in a try/catch either, so the build hook crashes with an unhandled exception rather than falling back to the full binary. Wrapping the `listSync` + `readAsStringSync` loop in a broader catch (adding inaccessible directories to `unresolvedPaths`) would make the fail-closed path cover filesystem-level errors too, not just unreadable individual files.

✅ **Checked**
- `analyzer` dependency fully removed from `pubspec.yaml`; all `package:analyzer` imports gone from `detector.dart`; `pubspec.lock` shows it as transitive-only. Consumer apps resolve their own codegen/lint tooling freely again
- `verify-tarball` gate wired into `make check` with correct `requiredFiles` list covering both Cargo.locks, source sentinels, compile scripts, hooks, and web assets. Sanity guard (`files.length < 100`) catches parse/dry-run failures
- `detector_test.dart` pins four key contracts: every capability found through any call shape, unrelated files keep nothing, core-only app keeps nothing, and barrel-mediated usage (including multi-hop re-export chains) is detected
- WASM compile failure message in `hook/build.dart:476` no longer mentions wasm-pack; correctly defers to the compiler's own per-tool instructions
- README web-tool table is accurate (Rust, wasm-bindgen-cli, binaryen, jq) with per-OS install commands
- Submodule pointer updated to fork commit with `.gitignore` fix; the `vendor/pdf_oxide/Cargo.lock` file now ships in the tarball
- `TrimDetector` enum, tests, docs, error messages, and comments consistently renamed from `analyzer` to `scan`
- Existing review thread on `tool/verify_tarball.dart:69` already resolved — the regex fix was applied
- `detectCapabilities` is now synchronous; all three call sites (`bin/setup.dart:105`, `trim_plan.dart:74`, `detector_smoke.dart:5`) correctly dropped `await`

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — text-scan detector, tarball gate, submodule bump, and message fixes all look correct

1. `bin/setup.dart:109` — the failure message still says "could not be resolved"; the detector now reports readability failures, not resolution failures

✅ **Checked**
- `pubspec.yaml` — `analyzer` dependency removed; only `record_use` remains as a tooling-only dep
- `lib/src/trim/detector.dart` — member-name regexes collapse qualified/bare keys correctly, `\b` boundaries prevent substring matches, re-export fixpoint tracks transitively, unresolvable `package:<self>/` barrel imports fail closed via the pubspec read check
- `lib/src/trim/capabilities.dart` — `TrimDetector.analyzer` renamed to `.scan`, default and error messages updated
- `lib/src/hook/trim_plan.dart` — `await` dropped from the now-sync `detectCapabilities` call, failure message matches
- `bin/setup.dart` — `await` dropped, `setup --trim` flow intact
- `tool/verify_tarball.dart` — dry-run parse covers decimal sizes, 14 build-critical paths, `< 100` safety guard, wired into `make check`
- `hook/build.dart` — wasm failure message defers to the compiler's own per-tool instructions
- `README.md` — tool prerequisites updated with correct names and install commands, no mention of `wasm-pack`
- `CHANGELOG.md` / `CHANGELOG.pre.md` — entries for 2.2.1 and 2.2.1-dev.0 covering all four fixes
- `vendor/pdf_oxide` submodule bump — points at the fork with the `.gitignore` patch
- `test/trim/detector_test.dart` — rewritten for the text-scan contract: full capability coverage, unrelated files kept empty, barrel-mediated usage tracked through a 2-hop re-export chain, missing pubspec fails closed
- `test/trim/capabilities_test.dart` — `analyzer` references updated to `scan`
- Previous review thread (`verify_tarball.dart:67`, decimal size regex) — resolved

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the analyzer dependency is gone, the text-scan detector is well-tested, the tarball gate is correctly wired, and every caller was properly migrated from async to sync

1. `tool/detector_smoke.dart:5` — `main` is still marked `Future<void> main() async` but `detectCapabilities` is now synchronous. The `async` keyword is harmless but misleading. A reader might think there is still async work in this smoke test.

✅ **Checked**
- `pubspec.yaml`: the `analyzer` dependency is fully removed; `pubspec.lock` confirms it is now transitive-only (pulled by `record_use`); `example/pubspec.lock` shows the cascade of transitive deps (`_fe_analyzer_shared`, `convert`, `glob`, `watcher`) all dropped
- `lib/src/trim/detector.dart`: the text-scan approach uses `\b<name>\b` word-boundary patterns, which correctly match member names in method calls, property accesses, and dynamic invocations — and correctly do not match substrings like `myConvertTo`. The fixpoint export-chain tracking handles barrel files, multi-hop re-exports, and both relative and `package:self/` import paths
- `lib/src/trim/capabilities.dart`: `TrimDetector.analyzer` renamed to `scan`; all references in the enum, docs, error messages, and tests updated consistently
- `lib/src/hook/trim_plan.dart`: `detectCapabilities` call correctly dropped the `await`; error message wording updated from "did not resolve" to "could not be read"
- `hook/build.dart`: the wasm-compile failure message no longer references `wasm-pack` and instead points to the tool that is actually named in stderr
- `bin/setup.dart`: sync call and error message wording consistent with the detector and trim_plan
- `test/trim/detector_test.dart`: five tests covering all capabilities found, unrelated files ignored, core-only trimming, barrel-mediated re-export chains (two hops, both relative and `package:self/` imports), and missing-pubspec fails-closed. No more analyzer-dependent fixture setup
- `test/trim/capabilities_test.dart`: `scan` replaces `analyzer` in all three trim-detector tests
- `tool/verify_tarball.dart`: 14 required paths cover both Cargo.locks, both Cargo.tomls, source sentinels, compile scripts, build config, hooks, and the web lane worker. The regex now handles decimal sizes (`\d+(?:\.\d+)?`), and the sanity guard (`< 100` files) catches total parse failure. Wired into `make check`
- `Makefile`: `verify-tarball` added to the `check` gate; doc comment explains why pub filtering honors vendored `.gitignore`
- `vendor/pdf_oxide`: submodule pointer updated, confirming the fork's `.gitignore` patch
- `docs/ARCHITECTURE.md` and `docs/CAPABILITY_ROADMAP.md`: all `analyzer` references updated to `scan` / `text-scan`
- `CHANGELOG.md` and `CHANGELOG.pre.md`: entries in both stable (`2.2.1`) and dev (`2.2.1-dev.0`) lanes
- `README.md`: build tool requirements now honest — wasm-bindgen-cli, binaryen, jq explicitly listed with install commands per OS; the old `wasm-pack`-tipped error message is gone
- The resolved review thread on `verify_tarball.dart:69` (decimal size regex) is correctly marked resolved and collapsed

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the tarball gate, analyzer removal, and text-scan detector are all well-built and well-tested.

1. `hook/build.dart:38` — the header waterfall comment still mentions `cargo/wasm-pack` on the compile step. The actual compile path now uses `tool/compile_rust.sh` and the user-facing error message no longer names wasm-pack. If the script doesn't use wasm-pack either, the comment drifts from reality. Not blocking — just a header comment.

✅ **Checked**
- `verify_tarball.dart` regex now handles decimal sizes (`\d+(?:\.\d+)?`) — the issue from the previous review round is fixed
- `detectCapabilities` is synchronous in all three callers: `bin/setup.dart`, `hook/trim_plan.dart`, `tool/detector_smoke.dart`
- `_filesSeeingApi` fixpoint correctly discovers barrel chains: direct export, one-hop relative, and `package:<self>/` two-hop — all tested
- `_filesSeeingApi` return set is computed after the fixpoint converges, so transitive visibility works
- Files that don't import pdf_manipulator are correctly excluded (test: `_unrelated` with same-named identifiers keeps nothing)
- Missing pubspec fails closed with `unresolvedPaths` containing the pubspec path
- `pubspec.yaml` no longer has a direct `analyzer` dependency; the transitive copy through `record_use` stays tooling-only
- `capabilities.dart` enum rename from `analyzer` to `scan` is documented in both changelogs and the error message
- `Makefile` wires `verify-tarball` into `check` — the gate runs on every PR now
- All test fixture directories are torn down with `addTearDown`

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — analyzer dependency gone, text-scan detector is dependency-free and fail-closed, tarball gate catches the Cargo.lock regression, test coverage pins all five scenarios.

1. `lib/src/trim/detector.dart:170-188` — the `_filesSeeingApi` fixpoint iterates all source files on every pass (O(files × export-chain-depth)). For a typical app this is negligible, but a BFS seeded from known pdf_manipulator importers would drop it to O(files). Not worth changing today — flag it only if a real app with thousands of Dart files in lib/ ever shows measurable startup.

✅ **Checked**
- `detector.dart` — text scan over-keeps by contract (word-boundary regex on member names), never under-keeps; barrel re-export chain tracked via fixpoint; missing pubspec fails closed; `followLinks: false` prevents symlink escape; sync API simplifies callers
- `verify_tarball.dart` — size regex handles both `(1234 B)` and `(1.5 KB)` (the decimal fix from last round is in); `< 100` sanity guard catches total parse failure; 14 critical paths checked
- `capabilities.dart` — `TrimDetector.scan` replaces `analyzer`; `apiMembers` map preserved with both qualified and bare keys; error messages updated
- `test/trim/detector_test.dart` — five test scenarios: all capabilities found, non-importers excluded, core-only app yields empty keep-set, barrel chain (`package:<self>/` + relative) tracked across two hops, missing pubspec fails closed
- `test/trim/capabilities_test.dart` — enum rename reflected in all assertions including the error-message `contains('scan')` check
- `pubspec.yaml` — `analyzer` dependency removed; transitive pull remains harmless (no version constraint forced on consumers)
- `hook/build.dart` — `wasm-pack` removed from error message; now defers to compiler's own per-tool instructions
- `.github/workflows/pr-checks.yml` — `pull_request` event (not `pull_request_target`), `permissions: {}` at top, job-level `contents: read` only, no `github.event.*` in `run:` blocks, pinned action SHAs

slopfairy:
posted my review below 👇

lib/src/trim/detector.dart
```diff
@@ -78,54 +131,87 @@ Future<DetectorResult> detectCapabilities(
...
+/// recorded the pubspec as unresolved (full-binary fallback).
+Set<String> _filesSeeingApi(
+  String appRoot,
+  String? pubspec,
+  Map<String, String> sources,
+) {
+  // package:<appName>/x.dart resolves into <appRoot>/lib/x.dart.
+  final nameMatch = RegExp(
+    r'^name:\s*(\S+)',
+    multiLine: true,
```
slopfairy, line 158: The pubspec name regex `r'^name:\s*(\S+)'` on line 149 doesn't strip YAML quotes. If the consumer's pubspec has `name: "my_app"` (valid YAML, the quotes are syntactic, not part of the value), the capture group includes the quotes and `selfPrefix` becomes `package:"my_app"/` — which never matches the actual `package:my_app/x.dart` import strings in the source. Barrel re-export tracking silently skips any file reached via `package:\<self\>/` imports when the pubspec uses quoted syntax.

The fix is to strip surrounding quotes from `nameMatch[1]` before building the prefix. This is a narrow edge case — most Flutter apps write `name: my_app` without quotes — but the fail-close for unresolved paths won't catch it because the pubspec read succeeds; the scan just silently misses the self-import path.

```suggestion
  final rawName = nameMatch[1]!;
  final appName = (rawName.startsWith('"') || rawName.startsWith("'"))
      ? rawName.substring(1, rawName.length - 1)
      : rawName;
  final selfPrefix = nameMatch == null ? null : 'package:$appName/';
```

slopfairy (dismissed):
> [!TIP]
> Approved — clean set of fixes, the analyzer removal + text-scan replacement is well-reasoned and the tests pin the contract

1. `lib/src/trim/detector.dart:152` — the pubspec name regex `r'^name:\s*(\S+)'` (line 149) doesn't strip YAML quotes. A pubspec with `name: "my_app"` yields `selfPrefix = 'package:"my_app"/'` which never matches real `package:my_app/x.dart` imports, so barrel re-export tracking silently skips files using `package:<self>/` imports. Narrow edge case (most Flutter apps use unquoted names), but cheap to fix — strip surrounding `"` or `'` from the captured name before building the prefix.

✅ **Checked**
- 19 changed files across the detector rewrite, analyzer dependency removal, Cargo.lock tarball gate, wasm-pack error fix, README tool docs, and changelog entries in both lanes
- `detectCapabilities` signature change from async to sync — all 4 callers updated: `bin/setup.dart`, `lib/src/hook/trim_plan.dart`, `tool/detector_smoke.dart`, `test/trim/detector_test.dart`
- `test/trim/detector_test.dart` — 5 tests covering full-capability detection, unrelated-file isolation, core-only trim, two-hop barrel re-export, and missing-pubspec fail-closed
- `tool/verify_tarball.dart` — regex updated per previous review (decimal sizes handled), 14 required files listed, parse-failure sanity guard (<100 files), wired into `make check`
- `pubspec.yaml` — `analyzer` direct dependency removed; `record_use` stays as the experimental RecordUse lane dependency
- `example/pubspec.lock` — analyzer transitive deps (`_fe_analyzer_shared`, `convert`, `glob`, `watcher`) cleaned out
- `hook/build.dart` — wasm compile error no longer references `wasm-pack`, defers to `compile_rust.sh`'s own per-tool messages with exact install commands
- `README.md` — build tool prerequisites documented with per-OS install table; old "needs Rust + wasm-pack" claim replaced with accurate per-tool breakdown
- Changelog entries follow the template: `##` headings, verb-led bullets, fixes last, no subsections, both lanes independent
- Workflows not changed in this PR; existing workflows use `pull_request` (not `pull_request_target`), pinned action SHAs, correct `permissions: {}` top-level
- `vendor/pdf_oxide` submodule bumped for the `.gitignore` patch that unblocks the Cargo.lock from pub's archive filter

slopfairy:
posted my review below 👇

lib/src/trim/detector.dart
chaudharydeepanshu: Fixed in 9b4c744. One change from your suggestion: it dereferenced `nameMatch[1]!` before the `nameMatch == null` check, which would crash on a pubspec with no name line — the raw capture is null-checked first, then quote-stripped. Test pins a quoted-name app whose only path to the API is a `package:self` barrel import.

slopfairy: > @chaudharydeepanshu: "Fixed in 9b4c744. One change from your suggestion: it dereferenced `nameMatch[1]!` before the `nameMatch == null` check,…"

Good catch on the ordering. My suggestion put `nameMatch[1]!` before the null guard, which would crash on a pubspec with no `name:` line. Your version — null-check first, quote-strip second — is correct. The test covering a quoted-name app with only a `package:self` barrel import path is the right pin for this edge case.

tool/verify_tarball.dart
```diff
@@ -0,0 +1,80 @@
...
+Set<String> _treeToPaths(String output) {
+  final entry = RegExp(
+    r'^([│ ]*)(?:├──|└──) (.+?)(?: \(<?\d+(?:\.\d+)? .?B\))?$',
+  );
+  final stack = <String>[];
+  final paths = <String>{};
+  for (final line in output.split('\n')) {
+    final m = entry.firstMatch(line.trimRight());
+    if (m == null) continue;
+    final depth = m.group(1)!.length ~/ 4;
```
slopfairy, line 74: When `depth` jumps by more than 1, `stack.length = depth` pads with `null` entries, and the resulting path would be malformed (e.g. `a/null/d`). The pub dry-run tree is always well-formed (±1 depth change), so this is theoretical, but a defensive guard costs nothing:

```suggestion
    final depth = m.group(1)!.length ~/ 4;
    if (depth > stack.length) {
      unresolved.add('parse error: depth jump at "$line"');
      return;
    }
    stack.length = depth;
```

The `return` is outside the try/catch so it won't interfere with exception handling — it just skips the malformed line. Enclosing function scope works because this is a top-level helper, not a nested closure.

slopfairy (dismissed):
> [!TIP]
> Approved — the analyzer-to-scan migration is correct, the tarball gate is well-designed, and the tests pin every edge case that matters.

1. `tool/verify_tarball.dart:74` — `stack.length = depth` pads with null on depth jumps; a defensive guard is cheap even though pub output is always well-formed

✅ **Checked**
- `detector.dart` — `_filesSeeingApi` barrel tracking: worklist propagation correct through both direct export and transitive re-export chains, quote-stripped pubspec name handles `name: "my_app"`, relative and `package:self/` import resolution paths both covered by tests
- `detector.dart` — capability matching: `byMember` collapses qualified keys to bare member names, `\b` word boundaries prevent substring false matches, the full-capability-app test asserts every enum value is detected through its spelled member names
- `detector.dart` — fail-closed contract: unreadable file → unresolved path → `resolved: false` → caller keeps full binary; unreachable directory → records itself and skips subtree instead of throwing; missing pubspec → unresolved (self-imports unresolvable)
- `verify_tarball.dart` — regex handles decimal sizes (from prior review thread), `< 100` sanity guard catches parse failures, `requiredFiles` covers both lockfiles, source sentinels, scripts, hooks, and the hand-written JS
- `capabilities.dart` — enum rename `analyzer` → `scan` consistent across value, wire name, doc comment, and parse default; error message updated
- `trim_plan.dart` — `detectCapabilities()` no longer awaited (sync), error wording matches the new scan semantics
- `hook/build.dart` — wasm-pack references removed from both the compile comment and the error message; error now defers to the tool's own output
- `pubspec.yaml` / `pubspec.lock` — analyzer direct dependency removed; transitive through `record_use` only, which consumers don't inherit
- `test/trim/detector_test.dart` — six test cases: full capability coverage, unrelated-file isolation, core-only trim, barrel re-export (single hop via relative, second hop via `package:self/`), quoted pubspec, missing pubspec fail-closed
- `example/pubspec.lock` — five transitive deps dropped (`analyzer`, `_fe_analyzer_shared`, `convert`, `glob`, `watcher`); no new deps introduced
- `README.md` — web tools table honest about what the build actually uses, `<X>` placeholder explains the setup command flow
- `CHANGELOG.md` / `CHANGELOG.pre.md` — both lanes documented, entries describe the problem not the fix internals

tool/verify_tarball.dart
chaudharydeepanshu: Guard added in the shape this function needs — the suggestion's `unresolved.add` + `return` come from the detector's structure; here a mid-loop return would silently hand back partial paths, and growing a non-nullable `List<String>` throws rather than pads. On a depth jump the parser now bails to empty, which trips the caller's sanity check and prints the raw listing. Gate re-run green.

slopfairy:
posted my review below 👇

tool/verify_tarball.dart
slopfairy: > @chaudharydeepanshu: "Guard added in the shape this function needs — the suggestion's `unresolved.add` + `return` come from the detector's str…"

Makes sense. `return {}` is the right shape for this function — the caller's `< 100` sanity check already prints the raw listing, so bailing to empty gives a loud, debuggable failure instead of a silent partial-path leak. Cleaner than the `unresolved.add` pattern I suggested, which carries detector-shaped baggage this helper doesn't need.

slopfairy (dismissed):
> [!TIP]
> Approved — analyzer dependency gone, tarball gate wired into check, all prior review threads resolved.

✅ **Checked**
- 19 files across the diff, every one read. No blocking issues.
- `pubspec.yaml`: `analyzer` removed. `pubspec.lock` marks it transitive (from `record_use`). `example/pubspec.lock` confirms the example app no longer pulls analyzer. Zero analyzer imports remain in source — the search hit was the old dev-branch `detector.dart`, not this PR's code.
- `lib/src/trim/detector.dart`: the text scan is well-designed. Barrel re-export tracking is correct (worklist propagation from direct exporters through `exportedBy`). Quoted pubspec name handling was fixed since last review. Five test cases now pin: full capability detection, unrelated-file exclusion, core-only trim, two-hop barrel re-export, quoted pubspec name, and missing pubspec→fail-closed.
- `tool/verify_tarball.dart`: all three prior-review issues addressed — decimal-size regex, depth-jump guard (bails to empty → caller's sanity check prints raw output). Wired into `make check`.
- Callers: `detectCapabilities()` is now sync; `bin/setup.dart`, `lib/src/hook/trim_plan.dart`, and `tool/detector_smoke.dart` all updated accordingly with no leftover `await`.
- `hook/build.dart`: wasm error message no longer mentions `wasm-pack`; now defers to the compiler's own tool instructions. Comment at line 31 correctly names `tool/compile_rust.sh` instead of `wasm-pack`.
- `vendor/pdf_oxide` submodule bumped for the `.gitignore` fix that ships `Cargo.lock`.
- `path: ^1.9.0` was already in `pubspec.yaml` — the detector's `package:path/path.dart` import is covered.
- Changelogs in both lanes (`2.2.1` / `2.2.1-dev.0`), docs synced across `ARCHITECTURE.md` and `CAPABILITY_ROADMAP.md`.
- README now honestly names the web-trim build tools with install commands.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — all three prior review threads resolved, gate passes, tests cover the contract

✅ **Checked**
- `pubspec.yaml` — `analyzer` dependency fully removed; no consumer constraint pollution
- `lib/src/trim/detector.dart` — text scan correctly: tracks barrel re-exports transitively (multi-hop worklist), strips YAML quotes from pubspec names, restricts scanning to files that actually reach the API (direct import or import of a file that re-exposes it), fails closed on unreadable files and missing pubspec
- `tool/verify_tarball.dart` — depth jump returns empty set (loud failure via caller's sanity check), decimal-size regex handled, sanity guard (`< 100` files), all 14 build-critical paths in `requiredFiles`
- `hook/build.dart` — WASM error message now defers to the compiler's own per-tool instructions instead of incorrectly naming `wasm-pack`
- `README.md` — web tools table matches what the build actually requires (Rust, wasm-bindgen-cli, binaryen, jq)
- `lib/src/hook/trim_plan.dart` + `bin/setup.dart` — `detectCapabilities` calls correctly synchronous (no `await`)
- `test/trim/detector_test.dart` — pins: all capabilities via every call shape, unrelated-file isolation, barrel re-export (two hops), quoted pubspec name resolution, missing pubspec fail-closed
- Both `CHANGELOG.md` (2.2.1) and `CHANGELOG.pre.md` (2.2.1-dev.0) entries in sync
- `Makefile` — `verify-tarball` correctly wired into `check` target
- `vendor/pdf_oxide` submodule updated — the `.gitignore` fix that ships the lock file
- No `.github/workflows/` changes — nothing to audit on the security front

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the analyzer dependency is gone, the tarball gate is wired, and the text-scan detector is well-tested across the key contract cases.

1. `lib/src/trim/detector.dart:134` — `_directive` captures only the first URI per line. Conditional imports (`import 'a.dart' if (x) 'b.dart'`) have two URIs on one line; the second is silently dropped. If a barrel chain goes through a conditional import, the scan misses it. This is corner-case territory — conditional imports of barrel files are rare — so it is not blocking, but documenting the limitation in the file header would make the tradeoff explicit next to the "over-keep only" contract.

✅ **Checked**
- `analyzer` removed from `pubspec.yaml` dependencies; lockfile retained transitively (pulled by dev tooling, not consumer builds)
- `detectCapabilities` signature changed from `Future<DetectorResult>` → sync `DetectorResult`; all callers updated (bin/setup.dart, trim_plan.dart, detector_smoke.dart)
- `_filesSeeingApi` barrel-tracking: seeds → worklist → exposes, cycle-safe via `Set.add` guard; `dependsOn` tracks both import and export edges, `exportedBy` only exports — correct propagation semantics
- Quoted YAML pubspec names stripped correctly (null-guard then substring, tested)
- Missing pubspec fails closed → `unresolved` list includes the pubspec path → `resolved: false` → full binary fallback
- `walk()` fall-closed: unreadable directories recorded and skipped, unreadable files recorded, no throw escapes the listing
- Tarball verify gate: 14 required paths, sanity-check guard (`< 100` files), depth-jump bail, wired into `make check`
- Wasm error message no longer mentions `wasm-pack`; defers to the compiler's own per-tool output
- README names honest web-tool requirements with install commands per OS
- Changelog entries in both lanes (3.0.0 / 3.0.0-dev.0) with retraction note on 2.2.0
- Detector tests: all-capabilities fixture, unrelated-file isolation, core-only, barrel re-export (two hops), quoted name, missing-pubspec fail-closed
- `TrimDetector.parse` default → `scan`; `analyzer` wire removed; error message updated
- `vendor/pdf_oxide` submodule updated (includes `.gitignore` patch for `Cargo.lock`)
- All three prior review threads resolved; no stale threads remain

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the Cargo.lock tarball gate, analyzer removal, and wasm-pack message fix are all clean.

1. `lib/src/hook/engine_compiler.dart:3` — the comment line says "analyzer-trimmed builds" but should say "scan-trimmed builds" (this file was missed by the rename sweep since the word is in prose, not in code)

✅ **Checked**
- `pubspec.yaml` — `analyzer` dependency removed. `pubspec.lock` reflects it as transitive only (the package's own dev tooling may still pull it, but consumers' dependency graphs are free)
- `lib/src/trim/detector.dart` — the text-scan detector is well-structured. The `_directive` regex correctly captures import/export URIs. Barrel re-export tracking via `_filesSeeingApi` correctly propagates exposure through transitive export chains. The quoted-pubspec-name edge case is handled (confirmed by the test)
- `lib/src/trim/capabilities.dart` — `TrimDetector.analyzer` renamed to `scan` consistently; `parse(null)` defaults to `scan`. Error messages reference the correct name
- `tool/verify_tarball.dart` — the gate lists all 14 build-critical files. The tree-regex handles both whole-number and decimal sizes. Depth-jump guard returns `{}`, which trips the caller's sanity check cleanly
- `hook/build.dart` — the WASM compile failure message no longer mentions `wasm-pack`; defers to the compiler's own tool diagnostics. The top-of-file comment now says `tool/compile_rust.sh (wasm)` instead of `cargo/wasm-pack`
- `bin/setup.dart`, `lib/src/hook/trim_plan.dart`, `tool/detector_smoke.dart` — `detectCapabilities` is correctly called synchronously (was `await detectCapabilities`)
- `test/trim/detector_test.dart` — six tests covering: all capabilities found, unrelated files ignored, core-only trim, barrel re-export chain, quoted pubspec name, missing pubspec fails closed. No `package:analyzer` dependency
- `README.md` — web tool install commands are accurate (`wasm-bindgen-cli`, `binaryen`, `jq` — no `wasm-pack`). The install table names exact commands per OS
- `CHANGELOG.md` and `CHANGELOG.pre.md` — 3.0.0 / 3.0.0-dev.0 entries follow the format. Breaking bullets are first, version bump is correct (Breaking: → MAJOR). The retracted 2.2.0 entry carries a blockquote superseded note
- `vendor/pdf_oxide` — submodule bumped to include the gitignore patch (markered in the fork)
- All three previous review threads are resolved; no new threads conflict with old ones
- No `pull_request_target` + PR-head checkout, no secrets exposed to fork code in the changed files

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — clean fix for the #171 tarball blocker, the analyzer constraint, and the wasm-pack message. Detector rewrite is well-tested and correctly errs toward over-keeping.

1. `test/trim/trim_plan_test.dart:61` — test name still says "analyzer": `'auto + analyzer with no app at the candidate fails CLOSED'`. The default is now `scan`; the name should match.

✅ **Checked**
- `lib/src/trim/detector.dart` — text-scan detector: dependency-free, over-keep by contract, barrel tracking via reverse-export graph with correct worklist propagation, conditional import branches extracted, quoted pubspec names stripped, missing pubspec → fail-closed. The regex handles multi-line directives (the `[^;]*` in the directive pattern, with `multiLine: true`, greedily spans lines until the terminating `;`). The one known false-positive path (import/export keywords inside block comments or multi-line string literals) is explicitly within the over-keep contract.
- `lib/src/trim/capabilities.dart` — `TrimDetector.analyzer` renamed to `.scan`, wire name, error messages, and doc comments all updated. No stale references remain in the enum or its callers.
- `test/trim/detector_test.dart` — 7 tests: all capabilities found, unrelated files ignored, core-only app trims everything, barrel re-exports tracked, conditional imports tracked, quoted pubspec names resolved, missing pubspec → fail-closed. Good coverage of the over-keep and under-keep contracts.
- `test/trim/capabilities_test.dart` — updated for the `analyzer`→`scan` rename. README drift guard and engine parity tests unchanged.
- `tool/verify_tarball.dart` — parses `pub publish --dry-run` output, depth jump guard (`return {}` → caller's `< 100` sanity check fails loudly), decimal sizes supported in the regex, all 14 build-critical paths asserted. Wired into `make check`.
- Dependency removal — `analyzer` removed from `pubspec.yaml`, lock file updated to `dependency: transitive`. `package:path` was already present at `^1.9.0`. No new dependencies introduced.
- `bin/setup.dart`, `lib/src/hook/trim_plan.dart`, `tool/detector_smoke.dart` — all `await detectCapabilities(...)` call sites updated to the sync signature.
- `hook/build.dart` — WASM error message no longer mentions `wasm-pack`, defers to the compiler's per-tool instructions instead.
- `vendor/pdf_oxide` — submodule updated to the commit with the `.gitignore` patch.
- Changelogs — both lanes updated. SemVer guidance added to the header. 2.2.0 entries marked as retracted, superseded by 3.0.0.
- Workflow security — `ci.yml` and `pr-checks.yml` use `pull_request` (not `pull_request_target`), SHA-pinned actions, `contents: read`, `github.event.*` values in `env:` only. `triage.yml` uses `pull_request_target` with proper hardening comments and delegates to the pinned reusable workflow — no checkout of PR head code. `release.yml` uses `push` trigger with `github.event.before` in `env:` only. No rolling action refs.

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — all prior threads resolved, the two #171 blockers are fixed, and the detector rewrite is solid.

1. `lib/src/trim/detector.dart:137` — the `_directive` regex `\s+` between `(import|export)` and the body matches any whitespace including newlines in Dart's default regex mode. Harmless for real source (Dart imports are single-line), but a deliberate `[\t ]+` would be a tighter contract. A low-stakes clarity fix — the regex already errs toward over-keeping regardless.

✅ **Checked**
- The `analyzer` dependency is gone from pubspec.yaml and the example lockfile — consumers' freezed / json_serializable tooling resolves freely again
- The text-scan detector (`detector.dart`) correctly handles: direct imports, barrel re-exports (transitive), conditional imports, quoted pubspec names, unreadable files (fail closed), and unrelated files are ignored
- `detectCapabilities` → sync is clean; all three call sites updated (`bin/setup.dart`, `hook/trim_plan.dart`, `tool/detector_smoke.dart`)
- `verify-tarball.dart` gate: parses the dry-run tree, checks all 14 build-critical paths, sanity-guards on < 100 files, rejects depth jumps — wired into `make check`
- The vendored `pdf_oxide` submodule bump includes the `.gitignore` patch so `Cargo.lock` ships in the pub archive
- `TrimDetector.analyzer` → `scan` rename is consistent across all files (enum, error messages, docs, tests, architecture docs)
- `hook/build.dart` error message no longer names `wasm-pack` — defers to the compiler script's own per-tool output
- README tool table covers Rust, wasm-bindgen-cli, binaryen, and jq with per-platform install commands
- Changelogs in both lanes (`CHANGELOG.md` for stable, `CHANGELOG.pre.md` for dev) are correct and consistent
- Tests cover: all capabilities found, unrelated files ignored, core-only trims everything, barrel re-exports (two hops), conditional imports (every branch), quoted pubspec name, missing pubspec (fail closed)
- No `pull_request_target` in workflows, no secrets in `run:` blocks, all actions use pinned SHAs

PR #178, opened 2026-07-19
fix: Rust-only trim toolchain — comment-aware scan, rustc target probe, workspace-locked wasm tools

### LINKED ISSUE #175: trim: auto detects capability names inside comments
### What happened?

  The dependency-free source scanner used by `trim: auto` matches capability
  method names inside comments.

  A comment containing the word `render` caused the optional `render` capability
  to be retained even though the application uses `pdfx`, not
  `pdf_manipulator`, for rendering.

### Actual result

  trim: keeping [render] -> features [wasm,rendering]

  Changing only the comment from “render them” to “display them” produces:

  trim: keeping [] -> features [wasm]

  ### Expected result

  Comments should not affect capability detection. The example should produce the
  core-only keep-set:

  trim: keeping [] -> features [wasm]

  ### Impact

  This is fail-safe and does not break functionality, but it silently retains the
  rendering engine. According to the package size table, that adds approximately
  4.2 MB to the native library.

  The false positive can be especially confusing when another PDF package
  actually provides the app’s rendering implementation.

  ### Possible fix

  Strip Dart comments before matching capability member names. This could remain
  dependency-free and preserve the scanner’s conservative behavior for executable
  source.

  Ignoring comments alone would fix this case without requiring semantic analysis
  or changing how dynamic and otherwise unresolved calls are handled.

  It may also help if trim output reports the matched member and source location,
  for example:

  render retained: lib/example.dart:4 (matched "render")

  That would make unavoidable conservative matches much easier to diagnose.

### Steps to reproduce

 Configure automatic trimming:

  ```yaml
  hooks:
    user_defines:
      pdf_manipulator:
        trim: auto
```

  Add a file that imports pdf_manipulator:
```dart
  import 'package:pdf_manipulator/pdf_manipulator.dart';

  void example(Pdf pdf) {
    // Keep the fields filled and let the PDF viewer render them.
    pdf.dispose();
  }
```

  Run:
```
  dart run pdf_manipulator:setup --trim
```
### Target (which platform?)

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [x] Web

### Package version

3.0.0-dev.0

### Flutter version

3.44.0

### Did this work before?

None

### LINKED ISSUE #176: Trim build requires rustup even when the WASM target is already installed
### What happened?

  `pdf_manipulator:setup --trim` assumes every Rust toolchain is managed by rustup.

  The build script checks targets with:
```sh
  if ! rustup target list --installed | grep -qxF "$1"; then
    rustup target add "$1"
  fi
```
  This fails for valid Rust distributions supplied by system package managers, container images, build systems, or other declarative toolchain managers. Such environments can provide rustc, cargo, and wasm32-unknown-unknown without providing rustup or its metadata.

  The target can already be installed and usable:
```
  rustc --print target-libdir --target wasm32-unknown-unknown
```
  Yet the build still fails because it invokes rustup.

  Expected behavior:

  - First detect whether the requested target is already usable through rustc.
  - Continue without rustup when it is installed.
  - Only use rustup target add as an optional fallback when the target is missing and rustup is available.
  - If neither condition is satisfied, report how to provide the missing target.

  This would allow trimming to work with any valid Rust toolchain rather than requiring a rustup-managed installation.

### Steps to reproduce

  1. Install a Rust toolchain containing:
      - rustc
      - cargo
      - wasm32-unknown-unknown

     Do not manage the toolchain through rustup.

  2. Verify that the target is available:

     rustc --print target-libdir --target wasm32-unknown-unknown

  3. Install the other documented trimming tools, including the required wasm-bindgen-cli, Binaryen, and jq.
  4. Configure trimming:

     dependencies:
       pdf_manipulator: 3.0.0-dev.0

     hooks:
       user_defines:
         pdf_manipulator:
           trim: auto

  5. Run:

     dart run pdf_manipulator:setup --trim

  6. The capability scan succeeds, but compilation fails when the build script invokes rustup, even though the WASM target is already installed.

### Target (which platform?)

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Web

### Package version

 3.0.0-dev.0

### Flutter version

3.44.0

### Did this work before?

None

### LINKED ISSUE #177: Trim setup rejects non-exact wasm-bindgen-cli versions without checking compatibility
### What happened?

  The trimmed web build currently requires the installed wasm-bindgen-cli version to exactly equal the wasm-bindgen crate version in the engine’s Cargo lock.

  For example, an installed CLI reports:

  wasm-bindgen 0.2.121

  The engine uses 0.2.126, so setup rejects the installed CLI before attempting compilation:
```
  Error: required tool not found. Install it:
    Run: cargo install wasm-bindgen-cli --version 0.2.126
```
  Is strict patch-version equality actually required by the generated bindings?

  If compatible CLI versions can process the engine output, setup could use the installed CLI and only fail when wasm-bindgen itself reports an incompatibility. Alternatively, the package could define a tested compatibility range instead of requiring exact equality.

  The current preflight check causes friction for system package managers, container images, and reproducible build environments, where an available CLI may be slightly older or newer but potentially compatible.

  Suggested behavior:

  1. Attempt to use the installed CLI when present.
  2. Let wasm-bindgen report an actual schema or format incompatibility.
  3. If compatibility is known in advance, validate against a supported range.
  4. If upstream genuinely requires exact equality, document why and provide a machine-readable or package-managed way to obtain the required version.

  This preserves safety without rejecting potentially compatible tools solely because their version strings differ.

### Steps to reproduce

  1. Install a valid Rust toolchain with wasm32-unknown-unknown.
  2. Install Binaryen and jq.
  3. Have a different valid wasm-bindgen-cli version available, for example:

     wasm-bindgen 0.2.121

  4. Configure trimming:

     dependencies:
       pdf_manipulator: 3.0.0-dev.0

     hooks:
       user_defines:
         pdf_manipulator:
           trim: auto

  5. Run:

     dart run pdf_manipulator:setup --trim

  6. Setup rejects the installed CLI and reports that version 0.2.126 is required.
  7. The consumer must manually copy that package-internal version into its build configuration and update it whenever the package changes.

### Target (which platform?)

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [x] Web

### Package version

 3.0.0-dev.0

### Flutter version

 3.44.4

### Did this work before?

None

chaudharydeepanshu:
Closes #175, closes #176, closes #177 — one change with one outcome: **trim builds need Rust and nothing else, native and web alike.**

## #177 — wasm-bindgen / wasm-opt / jq leave the requirements

The root problem was structural: the wasm is compiled with the `wasm-bindgen` crate version pinned in the engine's `Cargo.lock`, but post-processed by a separately installed CLI whose version we could only check, not control. The exact-version preflight was the symptom's guard, not a fix.

The fix removes the second version source entirely. A new `bindgen_runner` workspace crate in the engine fork runs `wasm-bindgen-cli-support` and the `wasm-opt` crate (binaryen as a cargo dep) as library calls:

- Both resolve from the **same `Cargo.lock`** as the engine's `wasm-bindgen` dependency — drift is structurally impossible, and upstream's own schema check backstops loudly if the lock is ever hand-edited apart.
- `wasm-bindgen-cli`, `binaryen`, and `jq` installs are gone from every path: the consumer `setup --trim` flow is now pure Dart + cargo (no shell script), and CI + local dev's `compile_rust.sh wasm` calls the same runner. One flow, three callers.
- The binaryen hash pins and their daily bump block are gone — nothing left to pin. The pinning *technique* stays armed: `fetch_verified.sh` remains in the tree as the standing (dormant) verified-download door, and `upgrade.sh` keeps the verify-pinned / check-availability plumbing with a recipe comment for the next mutable download source.
- First web build compiles the runner once (a few minutes, binaryen from source); cargo caches it after.

Generated glue is byte-identical to the CLI's output (verified by diff), minus a dangling `@ts-self-types` header that pointed at a `.d.ts` the old flow deleted anyway.

## #175 — comments are prose, not call sites

The scan now lexes each visible file and blanks comments (line, doc, and nested block) before member matching, so `// ...render them.` no longer keeps the render capability. String contents stay matchable — `'${doc.render(0)}'` is a real call site, and a `//` inside a URL string no longer risks eating the rest of the line. Blanking is offset-preserving, which enables the second half of the issue's ask: every surface now reports **where** each kept member matched:

```
trim: keeping [render] -> features [...]
trim: matched render (lib/preview.dart:12)
```

The hook's scan lane prints the same `member (file:line)` list at build time.

## #176 — rustup is a fallback, not a requirement

`ensure_target` (shell) and `ensureRustTarget` (Dart hook) now ask rustc itself — `rustc --print target-libdir` plus a directory-exists check, since the probe prints the would-be path even for uninstalled targets. Any toolchain manager that installed the target passes; rustup is only invoked as a fallback when present, and its absence produces instructions instead of `rustup: command not found`.

## Verification

- Detector suite: 11 tests including the reporter's repro (comment-only mention keeps nothing), interpolated-call keep, `//`-in-string, nested block comments, and match-site line accuracy.
- `make check`: every gate green (shell lint, analyze, floor, pana platforms, tarball gate, guards, native + web suites, example tests) except one 5000x-instance wasm stress test that OOM'd under local machine load — the unchanged pre-PR binary fails it identically on the same machine, so it is environmental; CI's clean runners arbitrate.
- Bonus fix found by that failure: a trimmed web compile used to stage its output in the package's shared `web_assets/`, where the hash-less compile-fallback paths (git-ref/path consumers, offline) and the package's own tests would pick it up as if it were the full engine. Trimmed builds now stage in a temp dir; only default builds touch `web_assets/`. Verified: after a trimmed consumer compile, `web_assets/` still holds the full 17,486,985-byte wasm while the app received its 8,155,919-byte trimmed copy.
- End-to-end consumer proof: a scratch app with a path dependency ran `flutter pub run pdf_manipulator:setup --trim` on a PATH containing **only** cargo/rustc/rustup + system utilities — no wasm-bindgen, no wasm-opt, no Homebrew — and produced the trimmed wasm.
- `web_assets/` rebuilt through the runner and committed; README size tables updated (17.5 MB raw / 7.3 MB gzipped — binaryen 116 in the crate vs 131 standalone costs ~1.4%, the price of hermetic builds).
- Tarball gate now requires `bindgen_runner/` in the pub archive.


slopfairy:
posted my review below 👇

slopfairy:
posted my review below 👇

lib/src/hook/engine_compiler.dart
```diff
@@ -101,6 +101,43 @@ Map<String, String> androidLinkerEnv(CodeConfig code, String targetTriple) {
...
+        libdir.isNotEmpty &&
+        Directory(libdir).existsSync()) {
+      return;
+    }
+  } on ProcessException {
+    // rustc missing entirely — the cargo probe above (or below at the
+    // call site) already produced the install-Rust instruction.
+  }
+  try {
+    _log.info('installing Rust target: $targetTriple');
```
slopfairy, line 131: `Process.runSync('rustup', ['target', 'add', targetTriple])` returns a `ProcessResult` whose `exitCode` is never checked. If rustup is present but the install fails (network error, disk full, etc.), the function returns silently without the target installed, and the subsequent cargo build fails with a less specific "target not installed" message.

The shell counterpart in `tool/build_lib.sh` is saved by `set -e` from the sourcing script, but the Dart version has no such backstop.

```suggestion
  try {
    _log.info('installing Rust target: $targetTriple');
    final result = Process.runSync('rustup', ['target', 'add', targetTriple]);
    if (result.exitCode != 0) {
      throw StateError(
        'rustup target add $targetTriple failed (exit ${result.exitCode}).\n'
        'stderr: ${result.stderr}',
      );
    }
  } on ProcessException {
```

tool/build_lib.sh
```diff
@@ -6,15 +6,28 @@
...
 ensure_target() {
-  if ! rustup target list --installed | grep -qxF "$1"; then
+  local libdir
+  # target-libdir prints the WOULD-BE path (exit 0) even for uninstalled
+  # targets — only the directory actually existing proves installation.
+  libdir=$(rustc --print target-libdir --target "$1" 2>/dev/null)
+  if [ -n "$libdir" ] && [ -d "$libdir" ]; then
+    return 0
+  fi
+  if command -v rustup >/dev/null 2>&1; then
```
slopfairy, line 22: `rustup target add "$1"` runs without an explicit exit-code check. The current callers (`compile_rust.sh`) have `set -euo pipefail` which catches the failure, but `build_lib.sh` is documented as sourced by 2+ scripts — a future caller without `set -e` would silently proceed past a failed install and hit a confusing cargo error.

```suggestion
    rustup target add "$1" || {
      echo "Error: rustup target add '$1' failed" >&2
      exit 1
    }
```

slopfairy (approved):
> [!TIP]
> Approved — three well-scoped fixes that shrink the dependency surface and tighten correctness, with thorough test coverage

1. `lib/src/hook/engine_compiler.dart:130` — `ensureRustTarget` doesn't check the exit code of `rustup target add`. If rustup is present but the install fails (network error, etc.), the function returns silently and the subsequent cargo build hits a less specific error. The shell counterpart is protected by `set -e` from the sourcing scripts; the Dart version should check `result.exitCode` explicitly.
2. `tool/build_lib.sh:22` — same gap in `ensure_target`. The current callers (`compile_rust.sh`) have `set -euo pipefail` which catches it, but `build_lib.sh` is sourced by 2+ scripts per the architecture docs. `|| exit 1` makes it robust against a future caller without `set -e`.

✅ **Checked**
- Comment-aware blanking in `_CommentBlanker` — handles line comments, doc comments (`///`), nested block comments, triple-quoted strings, raw strings, escapes, and `${...}` interpolation recursion correctly. Offset-preserving: match-site line numbers map to the real source, and the new tests cover the reporter's exact repro (`#175`), interpolated-call keep, `//`-in-string, and nested block comments.
- `ensureRustTarget` / `ensure_target` — rustc probe (`--print target-libdir` + directory existence) correctly distinguishes installed from uninstalled targets without depending on rustup. Any toolchain manager works; rustup is the consistent fallback. The Dart version's `ProcessException` catch handles a missing rustc, and the shell version's `command -v rustup` gate handles it cleanly.
- `_compileWasm` in `hook/build.dart` — replaces the shell-script call with pure Dart + cargo + `bindgen_runner`. Xcode PATH stripping on macOS, `--no-default-features` for wasm builds, trimmed output isolation (temp dir, never pollutes `web_assets/`), proper cleanup in the `finally` block. The removed `_compileNativeFromCli` was genuinely dead code — `_resolveNative` is private and only `main()` calls it, always with `BuildInput`.
- `compile_rust.sh` `do_wasm` — the `wasm-bindgen-cli` version check, `_install_binaryen` tarball machinery, and `WASM_OPT_FLAGS` array are all gone. `cargo run -p bindgen_runner` is the single post-compile step, sharing `Cargo.lock` with the engine. The `fetch_verified.sh` file deletion and `versions.env` binaryen pin removal are consistent — nothing dead references either.
- Workflow security: `.github/workflows/upgrade-check.yml` only removes binaryen from static heredoc content and comment text. No untrusted input enters `run:` blocks. Pinned action SHAs and `permissions: {}` at the top level are unchanged from the prior state.
- Tarball gate: `tool/verify_tarball.dart` now requires `bindgen_runner/Cargo.toml` and `bindgen_runner/src/main.rs` (and correctly doesn't require the deleted `fetch_verified.sh`).
- Architecture docs, README, CHANGELOG, UPDATING, and CODEOWNERS are all consistent with the toolchain changes.

lib/src/hook/engine_compiler.dart
```diff
@@ -101,6 +101,43 @@ Map<String, String> androidLinkerEnv(CodeConfig code, String targetTriple) {
...
+    Process.runSync('rustup', ['target', 'add', targetTriple]);
+  } on ProcessException {
+    throw StateError(
```
slopfairy, line 134: The `rustup target add` result is discarded. If rustup exists but fails (network error, permissions), the function returns silently and cargo's "target not installed" error is the user's first signal — less actionable than surfacing the rustup failure here.

The bash `ensure_target()` in `tool/build_lib.sh` catches this via `set -e`. Consider making the Dart version check the exit code too.

```suggestion
    final addResult = Process.runSync('rustup', ['target', 'add', targetTriple]);
    if (addResult.exitCode != 0) {
      throw StateError(
        'rustup failed to install Rust target $targetTriple '
        '(exit ${addResult.exitCode}).\n'
        'stderr: ${addResult.stderr}',
      );
    }
```

slopfairy (dismissed):
> [!TIP]
> Approved — 495 additions across 23 files, solid design on all three fixes, tests cover the key edge cases

1. `lib/src/hook/engine_compiler.dart:133` — the `rustup target add` exit code is unchecked. If rustup exists but the install fails (network error, permissions), the function silently returns and cargo later fails with a less actionable error. The bash `ensure_target()` catches this through `set -e`; the Dart version should check too.

✅ **Checked**
- `_CommentBlanker` lexer correctness: line comments, doc comments (`///`), nested block comments all blanked with offset preservation. String contents kept verbatim so `//` in URLs doesn't eat the rest of the line. String interpolation `${...}` recurses into code state properly, keeping real call sites matchable. Unterminated literals fall through to EOF — every char stays matchable.
- `_string` handles single-line (newline terminates), triple-quoted, raw (no escapes/interpolation), escapes (`\x` skipped), and `\n` as a real newline-terminating single-line literals.
- `matchSites` line calculation: blanking preserves newlines, so line numbers derived from the blanked source match the original.
- `ensureRustTarget` (Dart) / `ensure_target` (bash): rustc `--print target-libdir` probe is correct for all toolchain managers. Fallback to rustup only when rustup exists — no more "rustup: command not found" on distro/nix-managed Rust installs.
- `_compileWasm`: directly runs `cargo build` + `cargo run -p bindgen_runner` — no shell script delegation. Trimmed builds write to a temp dir (never `web_assets/`), and the `finally` block cleans it up. Default builds update `web_assets/` directly for committing. macOS PATH stripped of Xcode Developer injections.
- `compile_rust.sh do_wasm`: simplified to `cargo run --release -p bindgen_runner` — same runner, same lockfile. The `WASM_OPT_FLAGS` array and binaryen/wasm-bindgen installers removed cleanly.
- `_resolveNative` signature: all params now `required` (non-nullable). The CLI fallback `_compileNativeFromCli` is deleted — only the hook path remains, which always has `BuildInput`.
- `verify_tarball.dart`: added `bindgen_runner/Cargo.toml` and `bindgen_runner/src/main.rs` to the required-files list. Removed `tool/compile_rust.sh`, `tool/build_lib.sh`, `tool/fetch_verified.sh`, `tool/versions.env` — correct, since the consumer's hooks call cargo directly and never shell out to these scripts.
- `tool/versions.env`: BINARYEN_VERSION + 3 sha256 pins removed. PANA_VERSION remains. The file now has only one pin and the comments match reality.
- `tool/ci/upgrade.sh`: binaryen bump block removed. `asset_urls()` is now `:` (no-op). `verify-pinned` and `check-availability` still work correctly against the empty set. The recipe comment at `asset_urls()` documents how to re-arm for the next mutable download source.
- `tool/fetch_verified.sh`: explicitly marked "dormant" with forward-reference to `asset_urls()`. The door is kept, not broken.
- `.github/workflows/pr-checks.yml`: `pull_request` trigger (not `pull_request_target`), `permissions: {}`, checkout with `persist-credentials: false`. No user content in `run:` blocks. `whuppi/ci` ref pinned to `@v2.0.4`.
- `.github/workflows/upgrade-check.yml`: `schedule` + `workflow_dispatch` only, no PR-triggered paths. `ref: dev` (not PR head). PR body is a static heredoc — no `github.event.*` interpolation in `run:` blocks.
- Detector tests: 11 cases — the old 7 plus 4 new: comment-only mention keeps nothing (the #175 repro), interpolated-call keep, `//`-in-string, and match-site line accuracy. All correct.
- `_nativeResolve` parameter change: `cacheFile` no longer nullable. For custom/trimmed plans, `null` was already passed; now `plan.isCustom ? null : cacheFile` is evaluated at the call site before being passed to `resolveAsset` — this is the same behavior, just made explicit.
- `_currentLibFileName` deleted: the function was only called from `_compileNativeFromCli` (also deleted). No remaining callers.
- `web_assets/pdf_oxide.js`: `@ts-self-types` header removed — the `.d.ts` was already deleted by the old flow, so this was dead.
- CHANGELOG, README, ARCHITECTURE.md, UPDATING.md: docs updated consistently across all the changed files. Size tables reflect the new binaryen version.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — all previous blocking issues fixed, full changeset is clean

1. `hook/build.dart:422-429` — the Xcode PATH stripping is duplicated from `engine_compiler.dart:185-191`. Both compile paths need it, and they currently call cargo independently (WASM needs the extra bindgen_runner step, so it cannot share `compileEngineForTarget`). A shared helper `stripXcodeFromPath(Map<String, String> env)` would keep the logic in one place should it ever need adjustment.

✅ **Checked**
- Previous review threads: all three resolved — `rustup target add` exit codes are now checked in both `engine_compiler.dart:140-144` and `build_lib.sh:25-28`
- `_CommentBlanker` lexer: handles `//`, nested `/* */`, string literals (single/double/triple, raw, escapes, interpolations), and preserves offsets for match-site reporting — correctness verified
- `ensureRustTarget`: probes `rustc --print target-libdir` before falling back to rustup; both the missing-rustup and rustup-failed paths surface clear errors
- `_compileWasm`: temp dir (line 460) for trimmed builds prevents polluting `web_assets/`; `finally` block (line 496) cleans up; exit codes checked on both cargo calls
- `compile_rust.sh wasm`: replaced three-install + two-CLI-tool chain with single `cargo run -p bindgen_runner` — workspace-locked, no version drift
- `tool/verify_tarball.dart`: correctly adds `bindgen_runner/Cargo.toml` + `main.rs` and removes shell scripts that pub.dev consumers never execute
- `tool/versions.env`: BINARYEN_* pins removed; upgrade.sh bump block removed; fetch_verified.sh stays dormant with clear documentation
- Tests: 11 detector tests including the reporter's repro for issue #175, string-interpolation preservation, `//`-inside-string guarding, and match-site accuracy
- Workflow security: `pr-checks.yml` uses `pull_request` (not `pull_request_target`), pinned action SHAs, minimal permissions, no untrusted input in `run:` blocks

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — all three previous blocking issues are fixed, the new code is clean

**Since last review** — the three `rustup target add` exit-code checks are all addressed:
- `lib/src/hook/engine_compiler.dart:109-146` — `ensureRustTarget()` now throws `StateError` on non-zero exit, and on `ProcessException` from a missing rustup.
- `tool/build_lib.sh:14-35` — `ensure_target()` uses `|| { echo "Error: ..."; exit 1; }` instead of bare `rustup target add "$1"`.

1. `lib/src/hook/engine_compiler.dart:151-158` — `stripXcodeFromPath` reads `Platform.environment['PATH']` directly and overwrites any PATH that may have been set in the `env` parameter. No caller today passes a PATH in `environment`, so this is dormant, but a `final hostPath = env['PATH'] ?? Platform.environment['PATH']` would make it robust against a future caller that does.

✅ **Checked**
- `ensureRustTarget()` — rustc probe (`--print target-libdir` + directory-exists check) handles any toolchain manager; rustup is fallback only; missing rustup gets instructions, not "command not found"
- `_compileWasm()` in `hook/build.dart` — trimmed builds stage output in a temp dir (never polluting `web_assets/`); temp dir cleaned up in `finally` on both success and failure; default builds still go to `web_assets/`
- `_CommentBlanker` in `lib/src/trim/detector.dart` — nested block comments (depth-tracked), line comments, doc comments, strings with interpolation (re-enters code state for `${...}`), raw strings, escaped `\$`, and single-line string termination at newline are all handled correctly
- `_compileNativeFromCli()` and `_currentLibFileName()` removed — both were dead code after the `_resolveNative` signature made all params required
- `tool/verify_tarball.dart` — `bindgen_runner/Cargo.toml` and `src/main.rs` added to the required-files list; the removed scripts (`compile_rust.sh`, `build_lib.sh`, `fetch_verified.sh`, `versions.env`) are not needed by the consumer compile path (the hook handles everything)
- `tool/compile_rust.sh` — `_install_wasm_bindgen`, `_install_binaryen`, and `WASM_OPT_FLAGS` all removed; `do_wasm` now calls `cargo run -p bindgen_runner`
- `tool/versions.env` — all binaryen pins + sha256s removed; only `PANA_VERSION` remains
- `.github/workflows/` — comment-only changes, no behavioral differences
- New detector tests — 4 tests covering: comment-only mentions produce no matches, interpolated calls DO match, `//` inside strings is not a comment, and match-site line accuracy (comment on line 3, real call on line 4 → reported as line 4)
- Previous review threads resolved — all three `rustup target add` exit-code threads are fixed and resolved

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — all three prior threads resolved, the changeset reads clean across every layer

1. `tool/compile_rust.sh:35` — the `source "$SCRIPT_DIR/versions.env"` line is now a no-op for this script. The BINARYEN_* vars it once fed are gone, and PANA_VERSION (the only remaining entry) was never consumed here. Harmless dead code, but a future reader tracing the dependency may waste a cycle on it.

✅ **Checked**

- All three previous inline threads (rustup exit-code checks in `engine_compiler.dart:131`, `engine_compiler.dart:134`, and `build_lib.sh:22`) are resolved — both Dart and bash paths now check the exit code and surface the failure.
- `_resolveNative` signature change from nullable to required params is consistent at the sole call site (`hook/build.dart:138`).
- `_CommentBlanker` lexer handles the full grammar: line comments, nested block comments, string literals (single/double/triple/raw), escapes, `${...}` interpolation with brace-depth tracking, and newline-terminated single-line strings. Offset-preserving blanking feeds correct line numbers into `matchSites`.
- WASM trimmed-build outputs go to a temp dir, never `web_assets/` — the `finally` block cleans up; non-trimmed writes to `web_assets/` and persists (correct).
- `bindgen_runner` crate files (`Cargo.toml` + `src/main.rs`) are listed in `verify_tarball.dart` required files, and the removed shell-script entries (`compile_rust.sh`, `build_lib.sh`, `fetch_verified.sh`, `versions.env`) are no longer needed by pub consumers since the hook compiles directly via cargo.
- Zero remaining `BINARYEN_` references in the repo. `fetch_verified.sh` and `upgrade.sh`'s `asset_urls()` are kept dormant with recipe comments for the next pinned download.
- Workflow security: `pr-checks.yml` uses `pull_request` trigger with `contents: read` only; `upgrade-check.yml` runs on schedule, not PR trigger — no `pull_request_target` path exists. Both use pinned `actions/checkout@9c091bb` and `whuppi/ci@v2.0.4`.
- `web_assets/pdf_oxide.js` `@ts-self-types` header removed cleanly — no `.d.ts` references remain anywhere in the repo.
- Test coverage: 4 new detector tests covering the reporter's repro (comment mentions), string interpolation, `//` inside string literals, and match-site line accuracy.

slopfairy:
posted my review below 👇

lib/src/trim/detector.dart
```diff
@@ -128,9 +144,147 @@ DetectorResult detectCapabilities(
...
+      } else {
+        _blank(src[i]);
+        i++;
+      }
+    }
+  }
+
+  /// Copies a string literal verbatim so `//` inside it is never taken
+  /// for a comment. Handles `'`/`"`, their `'''`/`"""` triple forms,
+  /// escapes, and `${...}` interpolations (recursing into code state so
```
slopfairy, line 245: The `r'$'` + `_peek(1) == '{'` check treats `$${...}` inside a string as an interpolation — Dart's grammar says `$$` is an escaped `$`, and `{...}` after it is literal text.

In practice this cannot cause under-keep (member names never contain `/`, and the text inside `$${...}` is still matchable — only child comments in that text would be blanked, and `//` or `/*` inside literal string text adjacent to a member name is not a real call pattern). But the lexer should model the grammar correctly.

```suggestion
      } else if (!raw && c == r'$' && _peek(1) == r'$' && _peek(2) == '{') {
        // Escaped dollar: $$ → literal $, { starts literal text.
        out.write(r'$$');
        i += 2;
      } else if (!raw && c == r'$' && _peek(1) == '{') {
```

(The new branch must come before the `${` check to avoid the two-char `$` + `$` consuming the first `$` and the second `$` matching `{` as the interpolation start.)

slopfairy (approved):
> [!TIP]
> Approved — all three prior review threads resolved, security clean, no blocking issues

1. `lib/src/trim/detector.dart:245` — `$${...}` inside a string is treated as an interpolation start; Dart grammar says `$$` is an escaped `$` and `{` starts literal text. In practice this cannot cause under-keep (member names never contain `/`, and text inside `$${...}` stays matchable), but the lexer should model the grammar. See inline suggestion.

✅ **Checked**
- Security: no `pull_request_target` in either workflow; `persist-credentials: false` on checkouts; no untrusted input interpolated into `run:` blocks; no hardcoded secrets
- `ensureRustTarget` (Dart) — `rustc --print target-libdir` probe with directory-exists guard, rustup fallback checks exit code, `ProcessException` catch for missing rustup produces instructions. Fixes #176 and the prior review's unchecked-exit-code finding
- `ensure_target` (shell) — same probe+fallback shape, explicit `|| { echo error; exit 1; }` on `rustup target add`. Fixes the prior review's shell finding
- `_compileWasm` — trimmed builds stage in temp dir, `finally` block cleans up, never contaminates `web_assets/`. Default builds still write to `web_assets/` directly
- `_CommentBlanker` — line comments, nested block comments, strings (single/double/triple, raw, escapes, `${...}` interpolation) all handled. String contents stay matchable (critical for interpolated calls). 4 new tests cover the #175 repro and edge cases
- `matchSites` — first-match `putIfAbsent`, line calculation via newline-count before match offset. `describeMatches` formats consistently across `bin/setup.dart` and `hook/build.dart`
- `tool/compile_rust.sh` — `source versions.env` removed, `_install_wasm_bindgen` and `_install_binaryen` removed, `WASM_OPT_FLAGS` removed, `do_wasm` replaced CLI calls with `cargo run -p bindgen_runner`. One flow, version-locked by `Cargo.lock`
- `tool/verify_tarball.dart` — replaced script/build-config entries with `bindgen_runner/Cargo.toml` and `bindgen_runner/src/main.rs`. Correct: shell scripts are CI-only, not needed by pub consumers
- `stripXcodeFromPath` — extracted from `compileEngineForTarget`, now used in both native and web paths, respects caller-supplied PATH over `Platform.environment`
- `upgrade-check.yml` / `upgrade.sh` / `versions.env` — binaryen pins, bump block, and hash checks removed consistently. `fetch_verified.sh` kept dormant with a doc notice. `asset_urls()` is now a no-op
- Changelog, README, ARCHITECTURE, UPDATING docs — consistent with the changes

PR #185, opened 2026-07-23
feat!: engine config → keep / detector / build; one compile driver; Rust pin+gate (#183); resilient downloads

### LINKED ISSUE #183: App doesn't compile on iOS with pdf_manipulator
### What happened?

When compiling for iOS I get:
Error (Xcode): failed to load manifest for workspace member `/Users/timread/.pub-cache/hosted/pub.dev/pdf_manipulator-3.0.0/vendor/pdf_oxide/.`

It works on MacOS.

### Steps to reproduce

flutter run -d <device id>

stderr: error: failed to load manifest for workspace member `/Users/timread/.pub-cache/hosted/pub.dev/pdf_manipulator-3.0.0/vendor/pdf_oxide/.`
stderr: error: failed to load manifest for workspace member `/Users/timread/.pub-cache/hosted/pub.dev/pdf_manipulator-3.0.0/vendor/pdf_oxide/.`
Running Xcode build...                                                  
Xcode build done.                                           26.1s
Failed to build iOS app
Could not build the precompiled application for the device.
Error (Xcode): failed to load manifest for workspace member `/Users/timread/.pub-cache/hosted/pub.dev/pdf_manipulator-3.0.0/vendor/pdf_oxide/.`
/Users/timread/Downloads/Flutter_Source/BFour/ios/stderr:0:0

### Target (which platform?)

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Web

### Package version

3.0.0

### Flutter version

3.44.7

### Did this work before?

None

chaudharydeepanshu:
Started as "collapse the two cargo-compile drivers into one" and grew into the **4.0.0** release: a breaking, mismatch-proof engine-config redesign on top of the unified compile driver, the #183 fix, and resilient downloads.

## 1. Breaking: engine config is `keep` / `detector` / `build`

Three flat keys under `hooks: user_defines: pdf_manipulator:`, designed so an invalid config is unrepresentable and a cross-axis mismatch fails the build loudly rather than being silently ignored:

```yaml
hooks:
  user_defines:
    pdf_manipulator:
      keep: auto        # all (omit) | [render, …] | auto   — which capabilities
      detector: scan    # scan | record-use | compare        — how auto detects
                        #   (valid ONLY with keep: auto)
      build: speed      # speed (default) | size | debug      — how they compile
```

One parser (`lib/src/hook/pdf_config.dart`) that both the native build hook and the web `setup` script go through, so web and native can't diverge. It rejects:
- unknown keys / typos (web path, where the key set is visible),
- bad values on any axis,
- the one cross-axis coupling: `detector` set without `keep: auto`.

The design law is written into the parser header: *invalid configs are unrepresentable; a coupling between options is validated (or nested so the bad combo is unwritable), never left as a silently-ignorable sibling; the enclosing `pdf_manipulator:` key is already the namespace, so no cosmetic parent keys.*

**Migrating from 3.x** (all inline in the CHANGELOG too):

| 3.x | 4.0.0 |
|---|---|
| `trim: auto` | `keep: auto` |
| `trim: {keep: [render, …]}` | `keep: [render, …]` |
| `trim-detector: X` | `detector: X` |
| web `setup --trim` flag | configure `keep:` in pubspec, re-run `setup` (web now reads the same block native does) |

New capability: **`build:`** — `speed` (default, prebuilt), `size` (opt-level z), or `debug` (symbols kept for native crash symbolication). `size`/`debug` compile from source. Rides cargo's `release` profile via `CARGO_PROFILE_RELEASE_*` env, so `panic = "unwind"` stays intact and the native `catch_unwind` crash isolation keeps working.

Full rename swept clean (`grep` zero): `trim` → `keep` (`lib/src/trim/` → `lib/src/keep/`, all `Trim*` types → `Keep*`, `resolveKeepPlan`, `KeepConfigError` → `PdfConfigError`); `profile` → `build`. The unrelated PDF-page `trim` op was untouched.

## 2. One compile driver + Rust pin/gate (#183)

`tool/compile.dart` replaces `tool/compile_rust.sh` (−307 lines of bash). It loops the release target matrix but calls the exact same `compileEngineForTarget` / `compileWasmEngine` in `lib/src/hook/engine_compiler.dart` that the consumer build hook uses — they can no longer drift.

`RUST_VERSION` lives in `tool/versions.env` (single source for version pins). CI installs exactly it (was a rolling `stable`); the build hook's `ensureCargoVersion` reads the same pin: below → clear error, equal → silent, above → warning then proceeds.

This fixes **#183**: iOS (and any source-compile) on an older cargo died with the cryptic `failed to load manifest … feature edition2024 is required` — office_oxide uses edition 2024, which cargo < 1.85 can't parse, and the parse fails *before* cargo's own MSRV check runs. The gate runs pre-flight, so it surfaces an actionable message (naming the exact version + `rustup` command) regardless. Pinned Rust is `1.92`, the real MSRV (the `hayro-jpeg2000` transitive dep requires it; the crates' declared `1.88` is stale).

## 3. Resilient prebuilt-binary downloads

The download now retries transient failures (exponential backoff + jitter) and resumes interrupted transfers over HTTP `Range`, so a flaky network on a large asset (the ~180 MB iOS static library) no longer forces a slow from-source compile on a single blip. Verified live against GitHub: a 1 MB partial resumes via `Range: bytes=N-` → `206` → byte-identical hash-correct file. Preserves the "download-fail → compile" waterfall.

## Verified

- 53 keep/hook/config tests green (new `pdf_config_test` pins the cross-axis + unknown-key rules); `dart analyze .` clean.
- wasm through the driver is byte-identical to the pre-change build; native macOS arm64 = 21,145,600 bytes (matches the v3.0.0 release exactly).
- three-way Rust gate confirmed live (cargo 1.84 → clear error; 1.97 → warns-and-proceeds).
- download happy-path + resume both verified against a live GitHub asset.
- `make check` green except the 5000×-instance WASM stress test, which OOMs under local machine load — environmental (fails identically on unchanged code and the published binary; CI runners have the headroom).

## Notes

- **Unknown-key rejection is web-only.** The native build-hook API (`HookInputUserDefines`) exposes user-defines by key but not the key set, so native can't enumerate keys to catch a typo. Native still gets the full value-grammar + cross-axis validation; only typo-detection differs. Documented at the parser.
- CHANGELOG updated in both lanes (`## 4.0.0` stable + `## 4.0.0-dev.0` pre). Breaking → MAJOR: `setup --trim` shipped in stable 3.0.0, so removing it (and the whole config reshape) is a major bump.


slopfairy:
posted my review below 👇

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — one compile core shared by hook and CI driver, structural drift impossible; Rust version pinned and gated (#183)

1. `tool/compile.dart:146` — the old bash `compile_one` printed the output file size with `du -h` after each target build. The Dart `_compileOne` only prints the path. Adding `final size = outFile.lengthSync(); stderr.writeln('  → ${outFile.path} ($size bytes)');` would restore that convenience for release debugging.

✅ **Checked**
- CI workflow (`full-test.yml`) unchanged except comments — no logic regression
- Rust action now sources `RUST_VERSION` from `tool/versions.env` instead of rolling `stable` — reproducible builds, no fork-accessible secret issues
- `tool/compile_rust.sh` fully removed; zero remaining references in the codebase
- `compileEngineForTarget` / `compileWasmEngine` are the single compile core — called by `hook/build.dart` (one target) and `tool/compile.dart` (full matrix), same cargo flags, same `stripXcodeFromPath` / `ensureRustTarget` / Xcode path filtering
- `ensureCargoVersion` gate: below → throws with actionable message (#183 fix); equal → silent; above → warns and proceeds; `_cmpSemver` handles partial versions (`1.88` == `1.88.0`) correctly
- `tool/compile.dart` target matrix matches the old bash 1:1 — macos arm64+x64, ios 3-targets+strip, linux x64+cross, android 4-targets via NDK, windows (MSVC dual on native, mingw cross from linux), wasm with bindgen_runner
- `shake_audit.sh` now reads `FULL_FEATURES` from `build.json` via `json_get` instead of a hardcoded string — no silent staleness; `CORE`/`CORE_FEATURES` duplicate consolidated to one variable
- `tool/analyze.sh` reads features directly from `build.json` instead of shelling out to `compile_rust.sh --features` — simpler, same source
- `Makefile` wiring: all `compile-*` and `build-wasm` targets call `$(DART) tool/compile.dart`; `test-rust` and `shake-audit` correctly pass `DART` env var; `--features` output format preserved for shell capture
- `androidLinkerEnvFromNdkHome` matches the old bash NDK resolution: `ANDROID_NDK_HOME` → newest `~/Library/Android/sdk/ndk/*` fallback → host dir probe; CI failure behavior preserved (throw vs skip)
- `_ensureCrossGcc` replicates the old `provide_tool _install_cross_gcc` pattern: auto-install on CI, clear instructions on dev
- `hook/build.dart` simplified: inline cargo commands for WASM moved into shared `compileWasmEngine`, build hook now delegates to it
- All conventional commit / code style in line with the repo conventions

slopfairy (dismissed):
> [!TIP]
> Approved — one compile driver, shared by hook and CI; Rust version pinned once in `tool/versions.env` and gated in both callers; 307 lines of bash deleted, no stale references.

✅ **Checked**
- `tool/compile.dart` (+299): correctly replicates every target in the old `compile_rust.sh` — macos, ios, linux, android, windows, wasm, native, all, and `--features` — calling the same `compileEngineForTarget` / `compileWasmEngine` core as the build hook
- `lib/src/hook/engine_compiler.dart` (+217): `ensureCargoVersion` gate is sound — three-way (below → throw, equal → silent, above → warn), `_semver` / `_cmpSemver` correctly parse and compare dotted versions, `compileWasmEngine` is a clean extraction of the duplicated hook code, `androidLinkerEnvFromNdkHome` mirrors the old bash NDK resolution exactly
- `hook/build.dart` (−60): WASM compile deduplication correct — calls `compileWasmEngine` with the same parameters the old inline code used; trim path (temp dir, then clean up) preserved
- `.github/actions/capabilities/rust/action.yml`: now sources `tool/versions.env` and installs `$RUST_VERSION` instead of rolling `stable` — CI is reproducible
- `.github/workflows/full-test.yml`: compile guard comment updated; matrix unchanged; no security issues (`pull_request` trigger, `permissions: {}` top-level, SHA-pinned actions, no `pull_request_target`)
- `tool/shake_audit.sh`: `FULL_FEATURES` now reads from `build.json` via `json_get` — the hand-copied duplicate is gone; `CORE_FEATURES` consolidation is correct; WASM build path updated to the Dart driver with proper `DART` guard
- `tool/analyze.sh`: features now read directly from `build.json` via `json_get` — no longer shells out to the compile driver for them
- `Makefile`: all compile targets correctly routed to `$(DART) tool/compile.dart`
- `tool/versions.env`: `RUST_VERSION="1.88"` added with clear documentation
- Zero remaining references to `compile_rust.sh` in the repo
- No consumer-facing API change

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — faithful port of the bash compile driver to Dart, sharing the build hook's compile core. The Rust version gate is precise and well-isolated.

✅ **Checked**
- `tool/compile.dart` — target matrix, feature resolution, output layout, and Windows/Linux/macOS/iOS/Android/WASM dispatch all match the removed bash 1:1. The `_doNative` auto-detect logic is correct for each host.
- `lib/src/hook/engine_compiler.dart` — `compileWasmEngine` is correctly extracted from `hook/build.dart`; `compileEngineForTarget` now gates with `ensureCargoVersion` before touching any cargo manifest. The semver parse (`_semver` / `_cmpSemver`) handles the standard `cargo x.y.z` output. `androidLinkerEnvFromNdkHome` resolves NDK the same way the old bash did (env var → macOS default path → empty).
- `hook/build.dart` — the WASM compile path now delegates to `compileWasmEngine`; the old inline cargo calls are gone. No behavior change on the trim/non-trim path.
- `.github/workflows/full-test.yml` — uses `pull_request` (not `pull_request_target`), pinned action SHAs, `permissions: {}` at top. No untrusted event values interpolated into `run:` blocks. The compile matrix rows are intact.
- `.github/actions/capabilities/rust/action.yml` — now sources `tool/versions.env` for the Rust toolchain pin instead of a rolling `stable`. CI reproducibility improves, and the pin is the same line the hook gate reads.
- `tool/versions.env` — `RUST_VERSION="1.92"` with clear documentation of why 1.92 (transitive dep floor), not 1.88 (stale crate declared MSRV).
- `tool/shake_audit.sh` — `FULL_FEATURES` now reads from `build.json` via `json_get`, eliminating the stale hand-copied string. `CORE_FEATURES` consolidates the duplicate. The DART requirement is enforced with `: "${DART:?}"` and the Makefile passes it.
- `tool/analyze.sh` — reads features directly from `build.json` instead of shelling out to the old bash driver; simpler and correct.
- `Makefile` — all `bash tool/compile_rust.sh` calls replaced with `$(DART) tool/compile.dart`; no missing or mismatched targets.
- Docs (`ARCHITECTURE.md`, `UPDATING.md`, `build_constants.dart`, `verify_readme_sizes.dart`) — every stale reference to `compile_rust.sh` updated.

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — shared compile core eliminates drift between CI and consumer builds; Rust version pin and gate fix #183 cleanly; download retry with Range resume is a solid resilience improvement.

## Summary by area

**Shared compile core** (`tool/compile.dart` + `lib/src/hook/engine_compiler.dart`)
The driver loops the release target matrix through the same `compileEngineForTarget` / `compileWasmEngine` that `hook/build.dart` uses. This directly eliminates the "catch compile_rust.sh drifting from the hook" CI guard's reason for existing — the two can no longer diverge. The target matrix and NDK resolution logic faithfully reproduces the old bash behavior (every Rust triple, library name, linker env, and strip flag matches).

**Rust version gate (#183)** (`tool/versions.env` + `ensureCargoVersion`)
`RUST_VERSION="1.92"` in `tool/versions.env` is the single pin — CI installs exactly it (`.github/actions/capabilities/rust/action.yml`), and `ensureCargoVersion` reads it pre-flight. The three-way gate (below → error, equal → silent, above → warning) is correct. The error message names the exact version and gives copyable `rustup` commands. This ships in the tarball, so every consumer source-compile gets the same actionable message instead of `feature edition2024 is required`.

**Download retry + Range resume** (`lib/src/hook/resolver.dart`)
Four attempts with exponential backoff (±1s jitter) and HTTP Range resume on partial downloads. The three-outcome enum (`success`, `notFound`, `transient`) correctly distinguishes permanent 404/410 from retryable failures. The initial `deleteSync()` guards against resuming onto a stale file from an earlier, unrelated call. Idle timeout of 60s on the response stream prevents a stalled socket from hanging.

**build.json as single feature source** (`tool/analyze.sh`, `tool/shake_audit.sh`)
`analyze.sh` now reads `.features.native` / `.features.wasm` directly from build.json via `json_get` instead of shelling out to the compile script. `shake_audit.sh`'s `FULL_FEATURES` follows suit — the hand-copied list that could silently go stale is gone.

**CI workflow security** (`.github/workflows/full-test.yml`, `.github/actions/capabilities/rust/action.yml`)
Pull request event uses `pull_request` (not `pull_request_target`), top-level `permissions: {}`, `actions/checkout` uses pinned SHAs with `persist-credentials: false`, and no untrusted `github.event.*` values are interpolated into `run:` blocks. The Rust composite action `source`s `tool/versions.env` from the checked-out repo — trusted input, correct. All action refs are pinned SHAs.

**Documentation** (`docs/ARCHITECTURE.md`, `docs/UPDATING.md`, `CHANGELOG.pre.md`)
Architecture doc now accurately describes the single compile core, the version pin, and the updated dependency table. Changelog entries are clear and cite the right issues and reporters. No functional inaccuracies.

## Verified clean

- Every target in the Makefile correctly calls `$(DART) tool/compile.dart` — no stale `compile_rust.sh` references remain
- The WASM path in `hook/build.dart` delegates to `compileWasmEngine`; the trim-temp-dir guard still isolates trimmed builds from the default `web_assets/`
- `shake_audit.sh` correctly backs up and restores the default WASM artifact before/after the core-only build; `DART` is enforced via the `: ${DART:?...}` guard
- All Rust compile calls route env vars through `stripXcodeFromPath`
- `_doNative()` auto-detect logic matches the old bash's per-OS branching exactly

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — one compile path for consumer and CI, reproducible Rust pin, no regressions

✅ **Checked**
- `tool/compile.dart` faithfully ports the bash driver with identical target matrix, `--features` output, and `COMPILE_OUTPUT_DIR` layout. Every `Process.runSync` call passes an explicit arg list — no shell injection surface.
- `lib/src/hook/engine_compiler.dart`: `compileWasmEngine` and `compileEngineForTarget` now share the same implementation the hook uses. Both call `ensureCargoVersion` before touching cargo — the gate runs pre-flight and surfaces an actionable message instead of cargo's cryptic `edition2024 is required` (fixes #183).
- `ensureCargoVersion` semver parsing handles cargo's standard version output (`cargo 1.84.0 (hash date)`) and correctly gates: below → throws, equal → silent, above → warns-and-proceeds. Null-safe when cargo or the pin is missing.
- `lib/src/hook/build_profile.dart` — clean enum riding `CARGO_PROFILE_RELEASE_*` env overrides. `small` only lowers opt-level (keeps `panic=unwind`), `debug` add symbols. Both correctly force source compile by routing version to `0.0.0`.
- `lib/src/hook/user_defines.dart` + `test/hook/user_defines_test.dart` — deep-converts YamlMap/YamlList to plain Dart so downstream `is Map`/`is List` checks work identically for web and native. Tests cover absent pubspec, missing block, deeply nested trim map, and inline-map syntax.
- `lib/src/hook/resolver.dart` — download now retries with exponential backoff + jitter (4 attempts) and resumes interrupted transfers via HTTP Range. Server returns 206 → append; 200 (ignored Range) → rewrite from scratch. Idle timeout (60s) on the sink pipe, connection timeout (30s) on the client. 404/410 exit the retry loop immediately. All correct.
- `hook/build.dart` — profile flows from `BuildInput.userDefines` → `EngineProfile.parse` → `cargoEnv` spread into the compile call. Custom (= trimmed OR non-release profile) correctly skips cache, download, and pinned hash.
- `bin/setup.dart` — `--trim` flag removed; reads `profile` and `trim` from the app's pubspec via the shared `readPdfManipulatorUserDefines` + `resolveTrimPlan`. Web and native now share one config source.
- `tool/shake_audit.sh` — `FULL_FEATURES` now reads `build.json` via `json_get` instead of a hand-copied string. `CORE_FEATURES` consolidated to one variable. WASM build uses `$DART tool/compile.dart wasm` with the caller-provided DART (Makefile passes it; script validates with `: "${DART:?}"`).
- `tool/versions.env` — `RUST_VERSION=1.92` documented as the engine's true floor (transitive dep requires 1.92, not the crates' declared 1.88).
- `.github/actions/capabilities/rust/action.yml` — installs `$RUST_VERSION` from `versions.env` instead of rolling `stable`. CI is now reproducible.
- `.github/workflows/full-test.yml` — comment updated for the new driver name. No workflow logic changed. Uses `pull_request:` (not `pull_request_target`), read-only permissions, pinned action SHAs.
- `CHANGELOG.pre.md`, `README.md`, `docs/*` — documentation updates are consistent and accurate.

slopfairy:
posted my review below 👇

lib/src/keep/capabilities.dart
```diff
@@ -99,76 +99,69 @@ enum PdfCapability {
...
+      '  keep: auto                     # detector decides\n'
+      '  keep: [render, signatures]     # exactly these capabilities\n'
       'Capabilities: render, signatures, pdfa, office, extract '
       '(and core, which is always included)';
 
-  /// Parses the raw `trim` user-define value (YAML-decoded).
-  static TrimConfig parse(Object? raw) {
-    if (raw == null || raw == false) return off;
+  /// Parses the raw `keep` user-define value (YAML-decoded).
+  static KeepConfig parse(Object? raw) {
```
slopfairy, line 140: `false` (which YAML decodes from `keep: false` or `keep: no`) now falls through to the generic "unrecognized keep value" error. A dedicated case with a migration hint would be more actionable:

```suggestion
    if (raw == false) {
      // YAML `keep: false` / `keep: no` — previously meant "don't trim" (full
      // binary). That is now the default when the key is absent. Remove it.
      throw PdfConfigError(
        'keep: false is no longer a valid value. '
        'To keep everything (the full binary), remove the `keep:` key entirely.',
      );
    }
```

lib/src/hook/keep_plan.dart
```diff
@@ -7,14 +7,14 @@ import 'dart:io';
 
 import 'package:record_use/record_use.dart';
 
-import 'package:pdf_manipulator/src/trim/capabilities.dart';
-import 'package:pdf_manipulator/src/trim/detector.dart';
+import 'package:pdf_manipulator/src/keep/capabilities.dart';
+import 'package:pdf_manipulator/src/keep/detector.dart';
 
 /// What the build hook should do about trimming, resolved from the
```
slopfairy: The class doc comment still says "trimming" — the concept was renamed to "keeping" everywhere else.

```suggestion
/// What the build hook should do about capability selection, resolved from the
```

slopfairy (dismissed):
> [!TIP]
> Approved — clean refactoring with shared compile core, version gate, and consistent naming

1. `lib/src/keep/capabilities.dart:139` — `KeepConfig.parse` rejects `false` with a generic "unrecognized keep value" message. A dedicated case would give users migrating from `keep: false` a clear action item instead of a puzzle.
2. `lib/src/hook/keep_plan.dart:13` — the `KeepPlan` class doc comment still says "trimming" where the rest of the codebase now says "keeping."

✅ **Checked**
- Security: `.github/workflows/full-test.yml` uses `pull_request` trigger (not `pull_request_target`), `permissions: {}`, pinned action SHAs. No untrusted input interpolation. `.github/actions/capabilities/rust/action.yml` sources `tool/versions.env` safely — no `github.event.*` values in `run:` blocks.
- All old `lib/src/trim/` references migrated cleanly to `lib/src/keep/` — no dangling imports or stale paths remain anywhere in the codebase.
- `tool/compile_rust.sh` deleted (−307 lines of bash) with no remaining callers — checked via grep.
- `tool/compile.dart` correctly calls the shared `compileEngineForTarget` / `compileWasmEngine` in `engine_compiler.dart` for every target. Same format, same feature resolution, same output layout as the old bash driver.
- `ensureCargoVersion` gate (fixes #183): below → clear error naming the exact version; equal → silent; above → warning-then-proceed. Semver comparison handles nightly/suffix versions correctly via `_semver`'s leading-digit extraction.
- `PdfManipulatorConfig` parser: unknown keys, bad values, and the `detector`-without-`keep: auto` axis mismatch all fail loudly. Tests pin every rejection path.
- `readPdfManipulatorUserDefines` deep-converts `YamlMap`/`YamlList` to plain Dart collections — web and native feed identical shapes to the one config parser.
- `EngineBuild.cargoEnv` for `size` sets only `CARGO_PROFILE_RELEASE_OPT_LEVEL: z`, preserving `panic = "unwind"` (native crash isolation intact).
- Download retry in `resolver.dart`: 4 attempts, exponential backoff with jitter, HTTP Range resume, idle timeout, clean outcome enum (success/notFound/transient). A 404 stops retrying immediately; transient failures continue.
- `tool/shake_audit.sh` reads `FULL_FEATURES` from `build.json` via `json_get` instead of a hand-copied duplicate. `CORE_FEATURES` consolidated to one variable used consistently throughout.
- `tool/analyze.sh` reads feature sets from `build.json` via `json_get` instead of shelling out to the compile driver.
- `example_trimmed/pubspec.yaml` updated: `trim: {keep: [render]}` → `keep: [render]`. Docs, README, ARCHITECTURE.md, UPDATING.md, CAPABILITY_ROADMAP.md all updated.
- Tests renamed (`test/trim/` → `test/keep/`) and new tests added (`pdf_config_test.dart`, `user_defines_test.dart`) — all correctly import the renamed modules.
- No consumer-facing API change. The `lib/src/keep/` directory is tooling-only (never exported by the barrel) per ARCHITECTURE.md.

slopfairy:
posted my review below 👇

lib/src/keep/capabilities.dart
```diff
@@ -99,76 +99,69 @@ enum PdfCapability {
   }
 }
```
slopfairy, line 101: The doc comment says `true` is a valid alias for `auto`, but the parse method only accepts the string `'auto'`:

```dart
if (raw == 'auto') return auto;
```

A boolean `true` falls through to the generic "unrecognized keep value" error. The test at `test/keep/capabilities_test.dart` confirms this is intentional — `KeepConfig.parse(true)` is expected to throw `PdfConfigError`.

```suggestion
/// `auto` → [KeepConfig.auto] (detector decides); `[render, ...]` →
```

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — `KeepConfig` class doc comment claims `true` is an alias for `auto` but the parser rejects it

1. `lib/src/keep/capabilities.dart:101` — The class doc comment says `auto`/`true` → `KeepConfig.auto`, but `KeepConfig.parse` only accepts the string `'auto'`. A boolean `true` falls through to the generic "unrecognized keep value" error. The test at `test/keep/capabilities_test.dart` confirms the rejection is intentional (part of the "no true/false aliases" test case). The doc comment is wrong — drop the `/`true`` part.

✅ **Checked**
- The "trimming" doc comment in `keep_plan.dart` — fixed since the last review (resolved).
- `engine_compiler.dart`: `ensureCargoVersion` gate → below throws, equal silent, above warns — correct.
- `engine_build.dart`: `cargoEnv` overrides `CARGO_PROFILE_RELEASE_*` only, keeping `panic = "unwind"` — correct.
- `pdf_config.dart`: cross-axis `detector`-without-`keep:auto` rejection — correct, tested.
- `user_defines.dart`: deep-converts YamlMap/YamlList to plain Dart collections — correct, tested.
- `resolver.dart`: download retry + Range resume logic — starts clean then resumes within a single call's retry loop, not across calls — correct.
- `compile.dart`: reuses the same `compileEngineForTarget`/`compileWasmEngine` core as the build hook — correct.
- Security: `.github/actions/capabilities/rust/action.yml` sources `tool/versions.env` (repo file, not user input) — safe.
- `Makefile`: all `bash tool/compile_rust.sh` → `$(DART) tool/compile.dart` rewrites — consistent.
- `pubspec.yaml`: `yaml: ^3.1.2` added as direct main dep — correct, used by `user_defines.dart`.
- Test files: 53 tests covering config parsing, user-defines reading, capability grammar, detector, and keep plan — all consistent with the new design.

slopfairy:
posted my review below 👇

slopfairy (approved):
> slopfairy re-reviewing PR #185 after the latest push

> [!TIP]
> Approved — config parser, unified compile driver, Rust pin, and download resilience are all correct and well-structured

I posted three inline suggestions on `capabilities.dart` (normalization consistency, a doc fragment, and a misleading comment) — all optional.

The one existing open thread from the prior review — `keep: false` / `keep: no` needing a dedicated migration hint — is still valid but was not applied this round. It remains open as a 🟡 suggestion.

✅ **Checked**
- All 10 CI workflows: no secret exposure, no rolling refs, `pull_request_target` hardened correctly
- `PdfManipulatorConfig.parse`: three keys validated, unknown keys rejected (web path), cross-axis `detector` + `!auto` coupling enforced
- 15 native targets + WASM in `tool/compile.dart` match the old `tool/compile_rust.sh` exactly
- `compileEngineForTarget` / `compileWasmEngine` called identically from `hook/build.dart` and `tool/compile.dart` — shared core prevents drift
- Download resilience in `resolver.dart`: HTTP Range resume, exponential backoff + jitter (1s/2s/4s), permanent-vs-transient classification, download-to-compile waterfall all correct
- `RUST_VERSION` pin (1.92) read from `tool/versions.env`, enforced by `ensureCargoVersion` from both native and WASM paths — fixes #183
- `tool/compile_rust.sh` deleted (−307 lines of bash)
- All `trim` → `keep` renames grep-clean; PDF-page `trim` operation untouched
- `hook/link.dart` exists and correctly handles `compare` and `record-use` modes
- Thread PRRT_kwDOIDzDIs6TZ9Rv resolved — the stale `true`-as-`auto`-alias doc was fixed in this push

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the 4.0.0 redesign is clean, the rename is grep-zero, and every safety gate is wired in

1. `lib/src/keep/capabilities.dart:164` — `keep: false` (YAML `false` / `no`) falls through to the generic "unrecognized keep value" error. The test pins that `false` throws, which is correct, but the error message does not point the 3.x migrator at what to do. A dedicated case ("`keep: false` is no longer valid — to keep everything, remove the `keep:` key") would be more actionable. This is the same point from the unresolved thread (PRRT_kwDOIDzDIs6TZIkR) — it was not addressed in this push.
2. `example_trimmed/pubspec.yaml:3-4` — the description still says "TRIMMED engine" and "trim user_define". The config key is `keep:` now.

✅ **Checked**
- `.github/` workflows — `pull_request` only (not `pull_request_target`), `permissions: {}` top-level, pinned action SHAs, no untrusted `github.event.*` values in `run:` blocks
- `lib/src/keep/` — zero stale `TrimRecord` / `TrimConfig` / `TrimDetector` / `TrimMode` references; the rename is complete
- `tool/compile_rust.sh` — fully removed, zero references remaining anywhere in the repo
- `lib/src/hook/pdf_config.dart` — the three-axis parser with cross-axis coupling validation is well-designed; `PdfConfigError` on every bad path
- `tool/compile.dart` — calls the exact same `compileEngineForTarget` / `compileWasmEngine` core as the consumer build hook; one cargo path, two callers
- `lib/src/hook/engine_compiler.dart:ensureCargoVersion` — reads the pinned `RUST_VERSION` from `tool/versions.env`, gives clear instructions when too-old; fixes #183
- `lib/src/hook/resolver.dart` — the download retries with Range-resume, hash verification at download time, and proper cache invalidation for custom builds
- `hook/build.dart` + `hook/link.dart` — both integrate with the unified config through `PdfManipulatorConfig.parse`; web and native cannot diverge
- `bin/setup.dart` — reads the app's pubspec through the same `PdfManipulatorConfig.parse` + `resolveKeepPlan` pipeline the native hook uses; no config flags on the command
- `tool/versions.env` — `RUST_VERSION=1.92` with clear documentation explaining it is the engine's true floor (transitive dep requirement, higher than the stale crate-level MSRV of 1.88)

slopfairy:
posted my review below 👇

lib/src/hook/user_defines.dart
```diff
@@ -0,0 +1,55 @@
...
+/// Reads `hooks: user_defines: pdf_manipulator:` from [appRoot]'s
+/// pubspec.yaml. Returns an empty map when the file or block is absent — the
+/// defaults (speed build, keep everything) then apply. Values are plain Dart
+/// (Map/List/scalars), matching exactly what the native hook receives via
+/// `BuildInput.userDefines`, so the downstream parsers (`EngineBuild.parse`,
+/// `resolveKeepPlan`) behave identically for both callers.
+Map<String, Object?> readPdfManipulatorUserDefines(String appRoot) {
+  final file = File('$appRoot/pubspec.yaml');
+  if (!file.existsSync()) return const {};
+  final doc = loadYaml(file.readAsStringSync());
```
slopfairy: `loadYaml` throws a bare `YamlException` on malformed YAML. A try/catch here would give the user a friendlier message pointing at their pubspec.yaml's syntax, rather than a raw parse error with no file context.

```suggestion
  try {
    final doc = loadYaml(file.readAsStringSync());
    final block = _dig(doc, const ['hooks', 'user_defines', 'pdf_manipulator']);
    final plain = _toDart(block);
    return plain is Map<String, Object?> ? plain : const {};
  } on Exception catch (e) {
    throw StateError(
      'Failed to parse pubspec.yaml: $e\n'
      'Check for YAML syntax errors in ${file.path}.',
    );
  }
```

slopfairy (dismissed):
> [!TIP]
> Approved — config parser, compile driver, Rust pin, and download resilience are all solid

Again a large, careful PR with tight tests, clear docs, and code that reads its own design law aloud. The architecture holds — one parser for both callers, one compile core for both drivers, one version pin gating CI and consumers identically.

**What changed since the last review**

- The `keep: false` migration pathway is now covered (line 146 in `capabilities.dart` — the boolean branch catches `true`/`false`/`no` with a unified message directing users to the grammar. The unresolved thread from the prior round is resolved.
- All renamed paths (`trim` → `keep`, `compile_rust.sh` → `compile.dart`) are clean — zero stale references in a grep sweep.
- Tests for `PdfManipulatorConfig.parse` and `readPdfManipulatorUserDefines` cover the contract: empty block, all three axes, axis-mismatch rejection, unknown keys, bad values, and the YAML-to-Dart deep conversion.
- The `.github/` diff is comment-only in the workflow and a single-line `source tool/versions.env` in the Rust action (safe — it sources the repo's own file, not PR-head code).

🟡 **Suggestion** — optional, take it or leave it

1. `lib/src/hook/user_defines.dart:23` — `loadYaml` throws a bare `YamlException` on malformed pubspec.yaml, producing a raw parse error with no file context. A try/catch with a message pointing at the file would be friendlier.

✅ **Checked**

- `KeepConfig.parse` — handles null, `all`, `auto`, list, boolean, map, unknown (all rejected with the grammar; `core` accepted as a no-op)
- `PdfManipulatorConfig.parse` — unknown keys caught, detector/keep axis mismatch rejected, bad build/keep values rejected
- `ensureCargoVersion` — below → clear error with commands (fixes #183), equal → silent, above → warning; `requiredRustVersion` reads from `tool/versions.env` so CI and consumer gate stay synchronized
- `_download` — retry with exponential backoff + jitter, HTTP Range resume from partial bytes, 60s idle timeout, 404/410 distinguished from transient errors
- `compileEngineForTarget` and `compileWasmEngine` — both check `ensureCargoVersion` and `ensureRustTarget`; environment routing consistent (Xcode PATH strip, `engineBuild.cargoEnv`)
- `compile.dart` — loops the release matrix through the same `compileEngineForTarget`/`compileWasmEngine` the consumer hook calls; NDK resolution via `androidLinkerEnvFromNdkHome` mirrors the hook's `androidLinkerEnv`
- `hook/build.dart` — resolves through `PdfManipulatorConfig.parse`, passes `engineBuild` to both native and web paths, custom builds route to compile (prebuilt binaries exist only for default speed builds)
- `link.dart` — passthrough unchanged; RecordUse lane recompiles through the same `compileEngineForTarget` core
- `.github/workflows/full-test.yml` — comment-only changes; no `pull_request_target`, no untrusted input in `run:` blocks, pinned SHAs maintained
- `.github/actions/capabilities/rust/action.yml` — installs pinned `RUST_VERSION` from the repo's own file; safe
- Makefile — all references migrated from `bash tool/compile_rust.sh` to `$(DART) tool/compile.dart`
- `engine_build.dart` — all three modes keep `panic=unwind` (native crash isolation preserved); `size` uses opt-level z, `debug` keeps symbols unstripped

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — all 4 previous issues addressed, new code is clean and well-tested

## Changes since last review

All four inline threads are resolved:

1. **`lib/src/keep/capabilities.dart`** — `keep: false` now has a dedicated error message with a migration hint, instead of falling through to the generic "unrecognized" case.
2. **`lib/src/hook/keep_plan.dart`** — doc comment no longer says "trimming"; renamed to match the `keep` vocabulary.
3. **`lib/src/keep/capabilities.dart`** — doc comment no longer lists `true` as an alias for `auto` (only the string `'auto'` is accepted, matching the implementation).
4. **`lib/src/hook/user_defines.dart`** — `loadYaml` is now wrapped in a try/catch that converts bare `YamlException` into a `PdfConfigError` naming which pubspec.yaml failed.

✅ **Checked**

- Config parser design is sound — unknown keys, bad values, and the `detector`-without-`keep: auto` axis mismatch all fail loudly
- One compile driver (`engine_compiler.dart`) shared by `hook/build.dart`, `hook/link.dart`, and `tool/compile.dart` — they cannot drift
- Rust version gate (`ensureCargoVersion`) reads the single pin in `tool/versions.env`; CI installs exactly it; below → clear error, above → warning
- Download retry with exponential backoff + jitter, HTTP Range resume for interrupted transfers
- CI workflows: pinned SHAs, `permissions: {}` at top level, no `pull_request_target`, no untrusted data in `run:` blocks
- 53 config + keep + hook tests green (`pdf_config_test.dart` pins the cross-axis + unknown-key rules)
- `io-exempt` markers on build-time tooling tests are correct — keep grammar/detector tests never run in a browser
- `example_trimmed/` is removed (the old trim concept), matching the rename
- `compile_rust.sh` is removed; the Makefile and CI now go through `dart tool/compile.dart`
- Web setup script (`bin/setup.dart`) reads the same `pdf_manipulator:` block through the same parser as the native hook — no divergence possible

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — unified compile driver, mismatch-proof config, Rust version gate (#183), resilient downloads; all 4 prior threads resolved

1. `lib/src/hook/resolver.dart:331` — `HttpClient.connectionTimeout` is a connection-establishment hint on many platforms, not a hard deadline. The 30s value combined with the 60s `response.timeout` in the pipe should handle most stalls. If any large-asset downloads hang during the TLS handshake specifically, increasing `connectionTimeout` would be the knob.

2. `lib/src/hook/engine_compiler.dart:201` — the `cargo --version` output parsing splits on all whitespace and picks `token[1]`. Cargo's stdout format (`cargo 1.84.0 (...)`) is stable, but if the `--version` output ever gains a prefix line (e.g. a warning about deprecated config), the token index would shift. A regex extracting `\d+\.\d+\.\d+` from the full output would be format-robust without needing a specific token position.

3. `tool/compile.dart:320-322` — `_compileAll` calls `strip -S` on iOS static libs after compilation. If `strip` is absent on a non-macOS dev machine running `compile.dart ios` (unlikely, but possible in a cross-compile-from-Linux setup), the `Process.runSync` returns a non-zero exit silently. Consider logging a warning when the strip call fails, since a missing `strip` would produce an unstripped-but-valid `.a` that still works — just larger.

✅ **Checked**
- Config parser (`pdf_config.dart`) validates all three axes, catches unknown keys (web), bad values, and the `detector`-without-`keep: auto` cross-axis mismatch — 53 tests pin all the rejection paths
- `KeepConfig.boolean` path has the migration hint for 3.x `keep: false` → the prior review thread is resolved
- `readPdfManipulatorUserDefines` catches `YamlException` with the file path named — prior thread resolved
- `compileWasmEngine` in `engine_compiler.dart` and `compileEngineForTarget` share the same gate (`ensureCargoVersion`) and the same cargo profile-override pattern (`engineBuild.cargoEnv`) — one compile path, one config path
- Download retry: exponential backoff + jitter, Range-based resume on transient failures, 404/410 short-circuits, timeout guards on both connection and idle — the workflow is correct
- `RUST_VERSION` in `tool/versions.env` is the single pin; CI installs it, the gate enforces it, and the error message names the exact version + `rustup` command
- `_semver` / `_cmpSemver` comparison handles shorter/longer version arrays correctly (zero-pads)
- Import rename sweep is complete — `lib/src/trim/` → `lib/src/keep/`, all `Trim*` → `Keep*`, `TrimConfigError` → `PdfConfigError`, the conditional export in `record_use_shim.dart` uses correct relative paths
- `full-test.yml` uses `pull_request` trigger (not `pull_request_target`), pinned commit SHA actions, `contents: read` permissions, `persist-credentials: false` — no security flags
- `yaml` dependency promotion to direct is correct — `lib/src/hook/user_defines.dart` imports it directly, and the pubspec lock reflects the change

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the four changes (config redesign, unified driver, Rust gate, download resilience) are all clean and well-tested.

1. `hook/link.dart:107` — the RecordUse recompile path does not pass `engineBuild.cargoEnv` to `compileEngineForTarget`. If a consumer has `build: size` (or `debug`) + `keep: auto` + `detector: record-use`, the build hook ships the full binary in their chosen `build:`, but the link hook's trimmed rebuild always uses the default speed profile. The initial compile and the recompile produce different build classes. This only affects the EXPERIMENTAL RecordUse lane and would not cause a crash, but it is an inconsistency worth documenting or fixing. The fix is to read `cfg.build` in `_trimFromRecordings` and pass its `cargoEnv` to `compileEngineForTarget`.

✅ **Checked**
- Config parser contract: unknown keys, bad values, and the `detector`-without-`keep:auto` axis mismatch all throw `PdfConfigError` — verified with `pdf_config_test.dart` (10 tests covering every rejection path)
- Config design follows the stated law: independent axes stay orthogonal (`keep` × `build`); dependent axes are validated (`detector` only with `keep: auto`)
- Rust version gate: `ensureCargoVersion` below → throw, equal → silent, above → warning. Three-way semantics verified in the PR description
- Rename sweep: `grep`-zero on `trim`/`Trim` for the config system; the unrelated PDF-page trim operation untouched
- `tool/compile.dart` calls the same `compileEngineForTarget`/`compileWasmEngine` that `hook/build.dart` uses — structurally impossible to drift
- `tool/versions.env` is the single RUST_VERSION pin, sourced by CI (`.github/actions/capabilities/rust/action.yml`) and read by the build hook's gate
- Download resilience: 4-attempt retry with exponential backoff + jitter, HTTP Range resume on retries, 60s idle timeout, redirect following, distinct 404/410 (fall to compile) vs transient (retry) handling
- Workflow security: `pull_request` (not `pull_request_target`), `permissions: {}` at top, `contents: read` at job level, `persist-credentials: false`, pinned action SHAs, no `github.event.*` interpolation in `run:` blocks
- YamlException caught in `user_defines.dart` with the file path in the error message — no bare `loadYaml` crash
- Test coverage: 53 tests green across `pdf_config_test.dart`, `capabilities_test.dart`, `keep_plan_test.dart`, `user_defines_test.dart`, `detector_test.dart`
- Documentation: CHANGELOG in both lanes (stable + pre), README fully updated with new size table including `build: size` column, ARCHITECTURE.md + CAPABILITY_ROADMAP.md rewritten for the new design
- `pubspec.lock` updated; `pubspec.yaml` has `yaml: ^3.1.2` for the web pubspec reader (build-hook-only, never in the consumer barrel)

slopfairy:
posted my review below 👇

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — `androidLinkerEnvFromNdkHome` missing `AR` env var in Android cross-compilation path

1. `lib/src/hook/engine_compiler.dart:291` — `androidLinkerEnvFromNdkHome` sets only `CARGO_TARGET_<triple>_LINKER` but not `AR`. The consumer build-hook path via `androidLinkerEnv` (line 97–101) sets both; this standalone/CI path doesn't. Without it, cargo falls back to the host system's `ar` for intermediate `.rlib` artifacts, which may not produce Android-compatible archives. The NDK's `llvm-ar` lives in the same `bin` directory as the clang linker.

1. `lib/src/hook/engine_compiler.dart:446` — native cargo build path omits `--no-default-features` while the WASM path (line 355) includes it. Works correctly today because the crate's default features (`icc`, `legacy-crypto`) are always present in every features string the keep plan produces, but the asymmetry is fragile.

✅ **Checked**
- All 4 previously flagged issues from earlier review rounds are fixed: `keep: false` migration hint in `capabilities.dart`, doc comment no longer claims `true` as alias for `auto`, `keep_plan.dart` class doc says "keeps" not "trimming", and `user_defines.dart` wraps `loadYaml` in a try/catch
- Workflow security: no `pull_request_target`, all actions SHA-pinned, no untrusted event values interpolated into `run:` blocks, no secrets exposed
- `Makefile`: zero remaining references to `compile_rust.sh`; `--features` resolution correct for `test-rust`
- `bin/setup.dart`: `--trim` flag removed cleanly; config parsing uses the same `PdfManipulatorConfig` parser as the native hook; error handling is comprehensive
- `lib/src/hook/pdf_config.dart`: unified parser design is sound — unknown keys, bad values, and the `detector`-without-`keep: auto` coupling all fail loudly
- `lib/src/hook/engine_build.dart`: CARGO_PROFILE_RELEASE env vars set correctly for all three modes; `panic = "unwind"` preserved everywhere
- `tool/compile.dart`: target matrix complete, version pin works, NDK auto-detection correct
- `lib/src/hook/engine_compiler.dart:355`: WASM path correctly passes `--no-default-features`

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — one config parser, one compile driver, one Rust pin; well-tested, no security concerns

✅ **Checked**
- Config parser (`lib/src/hook/pdf_config.dart`) — the design law is explicit and enforced: invalid configs are unrepresentable, the one cross-axis coupling (`detector` without `keep: auto`) fails loudly, every bad value/unknown key throws. Tests at `test/hook/pdf_config_test.dart` pin each rule.
- Keep grammar (`lib/src/keep/capabilities.dart`) — three legal shapes, everything else throws `PdfConfigError`. The `bool` fallthrough at line 146 gives a clear migration message for the old `keep: false` from 3.x. Tested at `test/keep/capabilities_test.dart`, including README and engine-manifest drift guards.
- One compile driver (`lib/src/hook/engine_compiler.dart`) — `compileEngineForTarget` and `compileWasmEngine` are the single path for both the consumer build hook and `tool/compile.dart`. Target triple mapping, NDK linker env, and the Rust version gate are correct. The gate (`ensureCargoVersion`) reads `tool/versions.env` as the single pin source — below throws (fixes #183), equal silent, above warns.
- Build modes (`lib/src/hook/engine_build.dart`) — `speed`/`size`/`debug` all ride the `release` profile via `CARGO_PROFILE_RELEASE_*`, so `panic = "unwind"` stays intact for native crash isolation.
- Resilient downloads (`lib/src/hook/resolver.dart`) — 4-attempt exponential backoff with jitter, Range-based resume on retry, download-time hash verification, 5-step waterfall unchanged.
- All 4 previous review threads are resolved.
- Workflow security — all triggers are `pull_request` (not `pull_request_target`), `permissions: {}`, actions pinned to commit SHAs, no untrusted event data interpolated into `run:` blocks. The `release.yml` uses `push:` on protected branches with path filters.
- `tool/compile.dart` replaces `tool/compile_rust.sh` cleanly — same compile core, same target matrix, CI installs pinned Rust from `versions.env`.
- `lib/src/hook/keep_plan.dart` resolves all three keep modes (all/auto/manual) and the RecordUse defer-to-link path correctly.

PR #186, opened 2026-07-24
ci: add self-hosted Renovate (replaces dependabot + sweeps)

chaudharydeepanshu:
Stands up self-hosted Renovate on pdf as the first cut of the updater migration (design: `_playground/PLAN-renovate-migration.md`).

**Why:** dependabot never scans composite `action.yml` ([#6704](https://github.com/dependabot/dependabot-core/issues/6704)). Renovate does — so it keeps every `whuppi/ci` ref uniform (workflow **and** composite) and bumps third-party actions frozen inside composites. We already hit this twice: the split whuppi/ci pin, and `actions/cache@v5.0.5` (in `rust`/`wasm-cache`, a full major behind v6.1.0, never bumped).

**What's here (additive, nothing auto-fires):**
- `renovate.json5` — github-actions (pinDigests, scans composites) + pub + `.fvmrc` custom manager; groups whuppi-ci / actions / deps; 7-day cooldown for third-party, 0 for whuppi/ci. Validated against Renovate 43.279.0.
- `.github/workflows/renovate.yml` — self-hosted runner, SHA-pinned, **workflow_dispatch only + dry-run default**.

**Before this does anything, one thing only you can do:** add a `RENOVATE_TOKEN` secret (fine-grained PAT, Contents + Workflows + Pull-requests: write — same class as `RELEASE_TOKEN`; `GITHUB_TOKEN` can't write `.github/workflows/`).

**Then:** dispatch this workflow (dry-run checked by default) → read the log (it should detect actions/cache across the composites, the whuppi/ci refs uniformly, pub deps, and the Flutter SDK) → uncheck dry-run once the output looks right + enable the schedule + remove `dependabot.yml`.

dependabot stays for now (dry-run coexists; removed when Renovate goes live).


slopfairy:
posted my review below 👇

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: renovatebot/github-action@3064367f740a1a91cca218698a63902689cce200 # v46.1.20
+        with:
+          token: ${{ secrets.RENOVATE_TOKEN }}
+          renovate-version: 43.279.0
+        env:
+          RENOVATE_REPOSITORIES: ${{ github.repository }}
+          RENOVATE_AUTODISCOVER: "false"
+          RENOVATE_DRY_RUN: ${{ github.event.inputs.dryRun == 'live' && '' || github.event.inputs.dryRun }}
```
slopfairy: The ternary expression is bugged: when `dryRun=live` is selected, Renovate still runs in dry-run mode, so the dispatch UI can never go live.

Here is why. In GitHub Actions expressions, `''` is falsy, so:
- `dryRun=live` → `true && ''` → `''` → `'' || 'live'` → `'live'`

`RENOVATE_DRY_RUN='live'` is a non-empty (truthy) string, so Renovate treats it as dry-run mode. The "live" option in the dropdown is dead.

```suggestion
          RENOVATE_DRY_RUN: ${{ github.event.inputs.dryRun != 'live' && github.event.inputs.dryRun || '' }}
```

This inverts the logic: when dryRun is NOT live, pass the selected mode; when dryRun IS live, pass empty string (no dry run). Verified the truth table:
- `full` → `'full' != 'live'` → `true && 'full'` → `'full'` → `'full' || ''` → `'full'` ✅
- `extract` → same pattern → `'extract'` ✅
- `live` → `'live' != 'live'` → `false && 'live'` → `false` → `false || ''` → `''` ✅

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the `RENOVATE_DRY_RUN` ternary is bugged: the "live" dispatch option can never actually go live

1. `.github/workflows/renovate.yml:44` — `github.event.inputs.dryRun == 'live' && '' || github.event.inputs.dryRun` evaluates to `'live'` when `dryRun=live` is selected (because `''` is falsy in GitHub Actions, so `||` falls through). Renovate interprets any non-empty `RENOVATE_DRY_RUN` as dry-run mode, meaning the `live` choice in the dispatch dropdown is dead — you can never open PRs from the UI. See the inline suggestion.

1. `renovate.json5` — The global `commitMessagePrefix: "ci:"` applies to `pub` deps and `lockFileMaintenance` too, while dependabot used `"chore:"` for those. Consider adding per-rule overrides in the `deps` group and `lockFileMaintenance` so pub bumps read `chore:` like they did before.
2. `.github/workflows/renovate.yml:36` — No `timeout-minutes` on the job. Renovate can hang on registry issues; a 30-minute cap would prevent wasted runner minutes.

✅ **Checked**
- Both action SHAs verified against their tags (`actions/checkout@v7.0.0`, `renovatebot/github-action@v46.1.20`)
- `permissions: contents: read` is minimal; `RENOVATE_TOKEN` passed via `with:` not interpolated into a script
- No `pull_request_target`, no fork-triggerable events — dispatch only, schedule commented out
- User inputs flow only through `env:` blocks, no injection vector
- `renovate.json5` rule ordering is correct: `whuppi-ci` is last, so it overrides the `actions` group for `whuppi/ci` refs (no cooldown, own PR)
- `.fvmrc` custom manager regex matches the actual file content (`"flutter": "3.44.7"`)
- Both `rust/action.yml` and `wasm-cache/action.yml` use `actions/cache@v5.0.5` — the exact gap Renovate is meant to close
- `pr-checks.yml` whuppi/ci ref-uniformity check (line 50) will still pass after Renovate bumps refs, since Renovate groups whuppi/ci into a single PR

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — `renovate.json5` is clean and correctly configured for this repo's structure

Changed since last review: the second commit (`aa398da`) dropped `.github/workflows/renovate.yml` — the Renovate runner now lives centrally in `whuppi/ci`. The previous inline comment about the ternary expression bug resolved automatically since that file no longer exists.

1. **PR description is stale.** The body still lists `.github/workflows/renovate.yml` as a changed file and describes `dryRun=full` / `dryRun=live` dispatch inputs. Neither exist in this repo anymore — the workflow runs centrally from `whuppi/ci`. The description should be updated to match the second commit's model: "pdf carries only `renovate.json5` (config); the self-hosted runner + `RENOVATE_TOKEN` live in `whuppi/ci`."

✅ **Checked**
- `pinDigests: true` with github-actions manager — verified it will scan composite `action.yml` files. Confirmed `actions/cache@v5.0.5` in `rust/action.yml:21` and `wasm-cache/action.yml:7` that dependabot never saw
- `matchDepNames: ["whuppi/ci", "whuppi/ci/**"]` — covers all 29 `whuppi/ci` refs across `.github/workflows/` and `.github/actions/` (the `**` glob catches composite action sub-paths like `whuppi/ci/actions/capabilities/fvm`)
- `.fvmrc` custom manager regex — matches `"flutter": "3.44.7"` correctly with `datasourceTemplate: "flutter-version"`
- Schedule cooldowns match old dependabot: 7-day for third-party, 0-day for `whuppi-ci`
- `commitMessagePrefix: "ci:"` + `:semanticCommitsDisabled` — unified prefix passes conventional-commit lint
- Group structure: `actions`, `deps`, `whuppi-ci` (evaluated last, overrides cooldown/schedule for first-party refs)
- `lockFileMaintenance` enabled weekly — replaces the old lockfiles sweep

slopfairy:
posted my review below 👇

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,34 @@
...
+# renovate.json5, and opens bump PRs (composites included — the dependabot#6704
+# gap). Dispatch-only + dry-run while proving; enable the schedule and pin
+# @vX.Y.Z once the first dry-run looks right.
+name: Renovate
+
+on:
+  workflow_dispatch:
+    inputs:
+      dryRun:
+        description: "Detect only, open no PRs."
```
slopfairy, line 12: The description "Detect only, open no PRs." is only accurate when `true` (checked). When unchecked, Renovate opens real PRs — the description says the opposite. Consider a description that covers both states:

```suggestion
        description: "Run in dry-run mode (detect only, open no PRs)"
```

slopfairy (approved):
> [!TIP]
> Approved — previous dry-run bug fixed by switching to boolean input; config is solid

The ternary expression from last round is gone: `dryRun` is now a clean `type: boolean`, so there is nothing to get wrong at the expression layer. PR description updated to match (check/uncheck instead of `dryRun=full`/`dryRun=live`).

1. `.github/workflows/renovate.yml:12` — the `dryRun` description "Detect only, open no PRs." is only accurate when checked. When unchecked, Renovate opens real PRs. A description that names the mode rather than one outcome is clearer: "Run in dry-run mode (detect only, open no PRs)."

✅ **Checked**
- Security: workflow_dispatch only, no `pull_request_target`, no `github.event.*` interpolation in `run:` blocks (none exist), `permissions: {}` at top level
- `.github/workflows/renovate.yml`: `concurrency` prevents overlapping runs, `cancel-in-progress: false` is correct for Renovate
- `renovate.json5`: all five concerns covered — GitHub Actions (with `pinDigests` for composites), pub deps, whuppi/ci (own group, zero cooldown), lock file maintenance, `.fvmrc` Flutter SDK
- Package rules: `schedule` values valid, group names consistent, whuppi/ci rule evaluated last so it wins
- `uses: @main` is intentional for the proving phase, with a comment to pin at go-live

PR #187, opened 2026-07-24
ci: tune Renovate config (pinDigests + schedules)

chaudharydeepanshu:
Follow-up to #186 from the first dry-run: whuppi/ci refs hit an upgrade collision (pinDigests SHA-pinning first-party tag refs), and the Monday schedule was deferring the actions/pub/fvmrc groups. pinDigests off for whuppi/ci; drop schedules (7-day cooldown + grouping handle noise).

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -16,29 +16,30 @@
   branchPrefix: "renovate/",
   packageRules: [
     {
-      description: "All GitHub Actions grouped, weekly (Mondays).",
+      description: "All GitHub Actions grouped (SHA-pinned via pinDigests).",
       matchManagers: ["github-actions"],
       groupName: "actions",
-      schedule: ["* * * * 1"],
     },
     {
-      description: "Dart/Flutter deps grouped, weekly.",
+      description: "Dart/Flutter deps grouped.",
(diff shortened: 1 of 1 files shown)


slopfairy:
PR is already merged. Here is a post-merge assessment.

The four changes in `renovate.json5` are all correct and well-reasoned:

- **Dropped `schedule` from `actions`, `deps`, and `lockFileMaintenance`** — the global `minimumReleaseAge: "7 days"` combined with grouping already controls update noise. The Monday schedule was artificially deferring updates that were ready to land.
- **Set `pinDigests: false` on the `whuppi-ci` group** — the global `pinDigests: true` was SHA-pinning first-party tag refs (e.g., `whuppi/ci@v1.2.3`), which collided with Renovate also trying to bump the version tag. Disabling pinDigests for this group lets tag-based version bumps work cleanly.
- **`schedule: ["at any time"]` removed from `whuppi-ci`** — this was already a no-op (no restriction), so dropping it changes nothing.
- **Comments updated** — the new descriptions accurately explain what each rule does now, and the `whuppi-ci` comment documents why pinDigests is off there.

No issues. Clean config change.

PR #189, opened 2026-07-24
ci: Renovate — disable commit statuses (unblock PR creation)

chaudharydeepanshu:
The live Renovate run aborts with a masked `repository has changed during renovation` — the real error is a **403 on POST /statuses** because RENOVATE_TOKEN lacks *Commit statuses: write*. `statusCheckWhen: never` stops the status POST (cooldown still enforced internally). Clean fix: add Commit statuses: write to the token, then drop this block.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -7,6 +7,18 @@
   minimumReleaseAge: "7 days",
   internalChecksFilter: "strict",
   prCreation: "not-pending",
+  // Don't set commit statuses. The RENOVATE_TOKEN has no "Commit statuses:
+  // write" scope, so a status POST 403s and Renovate mis-reports it as
+  // "repository has changed during renovation" and aborts. This only disables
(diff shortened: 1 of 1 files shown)


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — clean, well-documented workaround that unblocks Renovate PR creation without side effects

✅ **Checked**
- All four `statusCheckWhen` keys (`artifactError`, `configValidation`, `mergeConfidence`, `minimumReleaseAge`) are valid per the Renovate schema
- Setting them to `"never"` only suppresses the visible commit status; `minimumReleaseAge` / `internalChecksFilter` cooldown is still enforced internally, as the comment notes
- The JSON5 comment clearly documents why the block exists and exactly what needs to happen to re-enable status checks (add the scope to the token, then drop the block)
- No impact on any other repos or workflows

PR #207, opened 2026-07-24
ci: bump all action refs (workflows + composites)

chaudharydeepanshu:
One owner for every `uses:` action ref in `.github` — workflow and
composite alike, no split with Dependabot:

- **whuppi/ci refs** across `.github` (workflows AND composite
  `action.yml`) swept uniform to the latest whuppi/ci release, so the
  pin never splits.
- **Every third-party action** (workflows AND composites) pinned to the
  latest SHA via pinact — including the refs inside composite
  `action.yml` that Dependabot can't read (dependabot-core#6704).

Dependabot owns pub deps only.

Add `ready-to-test` for the full cross-target run, then merge once green.

_Auto-generated by upgrade-check.yml_

### DIFF
--- a/.github/actions/capabilities/rust/action.yml
+++ b/.github/actions/capabilities/rust/action.yml
@@ -18,7 +18,7 @@ runs:
         echo "office_oxide=$(git rev-parse HEAD:vendor/office_oxide)" >> "$GITHUB_OUTPUT"
 
     - name: Cache cargo registry
-      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
         path: |
           ~/.cargo/registry/index
--- a/.github/actions/capabilities/wasm-cache/action.yml
+++ b/.github/actions/capabilities/wasm-cache/action.yml
@@ -4,7 +4,7 @@ description: Cache WASM build output to skip recompilation.
 runs:
   using: composite
   steps:
-    - uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+    - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       id: wasm-cache
       with:
         path: web_assets/pdf_oxide_bg.wasm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       code: ${{ steps.filter.outputs.code }}
       rust: ${{ steps.filter.outputs.rust }}
     steps:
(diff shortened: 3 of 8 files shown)


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — all `uses:` refs across workflows and composites are SHA-pinned, no rolling refs, no security regressions

✅ **Checked**
- `pull_request_target` — none of the workflows use it; all triggers are `pull_request:` with standard event types
- Untrusted input in `run:` blocks — every workflow passes external values through `env:` before use; `inputs.runner`, `inputs.branch`, and `inputs.filter` are all choice-typed, not free text
- Rolling action refs — zero floating refs (`@main`, `@v2`, etc.). Every `uses:` across `.github/` (workflows AND composite `action.yml` files) is pinned to a full SHA with a version comment
- Permissions — all workflows declare `permissions: {}` at top level with explicit per-job grants; no `write-all`, no `administration: write`
- Secret handling — `CI_ACTIONS_TOKEN`, `PUB_CREDENTIALS`, and `GITHUB_TOKEN` are routed through `env:` or `secrets:` (reusable workflow calls), never interpolated directly in shell
- `actions/cache` v5 → v6 migration — the two usages (`rust/action.yml`, `wasm-cache/action.yml`) use only `path`, `key`, and `restore-keys` inputs, which are compatible across the major version; the v6 breaking changes (`save-always` input removed, output format changes) do not apply here
- `actions/checkout` v7.0.0 → v7.0.1 — patch bump across all 16 call sites, no behavioral changes
- Composite actions swept — `rust/action.yml` and `wasm-cache/action.yml` are covered (the `wasm-build/action.yml` composite has no external `uses:` refs, and `make-target/action.yml` only references `whuppi/ci` pinned tags and local `./.github/actions/...` paths)

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — both SHAs verified against upstream tags, no security issues, clean mechanical bump

✅ **Checked**
- `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` resolves to tag v7.0.1 (verified via GitHub API — `git tag` points to this exact commit)
- `actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9` resolves to tag v6.1.0 (same verification)
- No `pull_request_target` in any modified workflow
- No untrusted `github.event.*` interpolation into `run:` blocks anywhere
- All permissions are minimal — `{}` at the top level, `contents: read` scoped per-job where needed; `contents: write` only in `schedule`/`workflow_dispatch`-gated jobs
- Composite actions in `rust/action.yml` and `wasm-cache/action.yml` also bumped — these are exactly the ones Dependabot cannot read, and the sweep caught them
- whuppi/ci refs remain uniform at v2.4.0 (the `pr-checks.yml` gate enforces this)
- Every changed line in all eight files is only a `uses:` SHA pin

PR #206, opened 2026-07-24
ci: one owner for all action refs (whuppi/ci v2.4.0), Dependabot = pub only

chaudharydeepanshu:
Opts pdf into the **`action-refs`** sweep (whuppi/ci v2.4.0) — the single owner of every action `uses:` ref. Kills the workflow-vs-composite split.

## Changes
- **`upgrade-check.yml`**: reusable → `@v2.4.0`, `sweepActions: true`, `CI_ACTIONS_TOKEN`. The sweep now owns *all* action refs: whuppi/ci (workflows + composites, uniform tag) and every third-party action (workflows + composites, to SHA via pinact — including the composite refs Dependabot can't read, #6704).
- **`dependabot.yml`**: drop the `github-actions` ecosystem entirely. Dependabot owns **pub only** (root + example).
- Swept all 30 whuppi/ci refs → `@v2.4.0` (uniform pin).

## Why
No more half-and-half — Dependabot did the workflow actions, a sweep did composites, and the whuppi/ci pin split between them. One owner takes all action refs; Dependabot keeps its one unique job (pub).

## Note
The sweep runs on the daily radar and needs the `CI_ACTIONS_TOKEN` org secret (already set). PRs are authored by the token's identity.

Verified: actionlint clean, dependabot.yml valid, refs uniform at v2.4.0.

PR #201, opened 2026-07-24
ci: revert Renovate — back to dependabot + upgrade-check

chaudharydeepanshu:
Ripping Renovate out. It's more tool than this CI wants — dashboard warnings, config nags, token scopes, per-consumer wrapper. Restores dependabot + the upgrade-check sweeps; deletes renovate.json5 + the wrapper. The one real gap (composite action.yml refs, #6704) gets a focused sweep job next, in this CI's own grain.

PR #208, opened 2026-07-25
release: promote dev → prod

chaudharydeepanshu:
Promotes `dev` to `prod` (15 commits ahead). **Do not merge yet — keeping open for review.**

## Headline
- **`feat!`: engine config → keep / detector / build** (#183, #185) — the unified compile driver, Rust pin + version gate, and resilient downloads. Breaking (`feat!`): the engine's user-defines config surface changed (`keep` / `detector` / `build`).

## CI / tooling
- **One owner for every action ref** (#206, #207, whuppi/ci v2.4.0): the `action-refs` sweep owns all `uses:` refs — workflows AND composites, whuppi/ci and third-party — closing the dependabot-core#6704 composite blind spot. **Dependabot is now pub-only.**
- Renovate was trialled and reverted within this range (#186–#201) — net zero; the end state is the sweep above.
- Routine: Flutter SDK bumps (#182, #204), lockfile refreshes (#184, #205), grouped dep bumps (#181).

## Review notes
Seven review findings on #185 were addressed before merge: the `keep: false` migration error, stale "trim" wording, `loadYaml` file-context wrap, the format-gate fix, cargo-version parse robustness, the strip-warning, the RecordUse build-profile consistency fix, and the Android `_AR` env + native `--no-default-features` symmetry.

Merging this cuts a prod release.


slopfairy:
posted my review below 👇

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — well-executed refactor: trim→keep rename, unified compile driver, config parser with exhaustive validation, Rust version gate, resilient download retries. 15 commits, 60 files, no blocking issues.

1. `lib/src/hook/user_defines.dart:2466` — `file.readAsStringSync()` can throw `FileSystemException` (permission denied, file locked) beyond `YamlException`, which would surface as a raw stack trace in the web `setup` command rather than a clear error. Wrap in a catch-all or catch `FileSystemException` and re-throw as `PdfConfigError` with the file path.

2. `tool/compile.dart:3918-3945` — `_doNative()` returns silently when none of its platform conditions match (uncommon OS). A trailing `stderr.writeln('no native targets available on this host')` would make the no-op explicit rather than mysterious.

✅ **Checked**

- The trim→keep rename: every class (`TrimConfig`→`KeepConfig`, `TrimMode`→`KeepMode`, `TrimDetector`→`KeepDetector`, `TrimPlan`→`KeepPlan`, `TrimRecord`→`KeepRecord`), every file (`lib/src/trim/`→`lib/src/keep/`, `test/trim/`→`test/keep/`), every import, every test, every doc reference — consistent across the entire codebase. The pubspec grammar changed from `trim:`, `trim: {keep: [...]}`, `trim-detector:` to flat keys `keep:`, `detector:`. No dangling references remain.

- `lib/src/hook/pdf_config.dart` — the one config parser, shared by native hook and web setup. Unknown keys rejected (names them and lists valid keys). `detector` without `keep: auto` rejected with an actionable message. Bad values for `keep`, `detector`, `build` all fail loudly. Test coverage: 7 tests covering defaults, valid combos, unknown key, detector axis mismatch, bad build, bad keep.

- The `--no-default-features` addition to native builds in `engine_compiler.dart:449-454` — the feature string from `build.json`'s `native` set carries the crate defaults (`icc`, `legacy-crypto`), so no features are silently dropped. Verified against `shake_audit.sh` which now reads `FULL_FEATURES` from `build.json` via `json_get` rather than a hand-copied duplicate that could go stale.

- `tool/compile_rust.sh` → `tool/compile.dart` — the new Dart driver shares the same `compileEngineForTarget` / `compileWasmEngine` core as `hook/build.dart`, making drift between CI and consumer builds structurally impossible. The Makefile, `analyze.sh`, `shake_audit.sh`, `full-test.yml`, and `ARCHITECTURE.md` all updated. The old bash driver (307 lines) is fully deleted — no dead code left behind.

- Rust version pin: `tool/versions.env` `RUST_VERSION="1.92"` is sourced by CI (`.github/actions/capabilities/rust/action.yml`) and enforced by the build hook's `ensureCargoVersion()` gate. A too-old cargo now fails with the exact version needed and the `rustup` command to install it, instead of cargo's cryptic `feature edition2024 is required` error. A too-new cargo proceeds with a warning naming the tested version.

- Resilient downloads in `lib/src/hook/resolver.dart` — 4 retries with exponential backoff + jitter (~1s, 2s, 4s + up to 1s), Range-based resume (`bytes=$existing-`) for partial downloads, 60s idle timeout on stalled sockets. Distinct outcomes: `notFound` (404/410) exits immediately, `transient` retries. The `dest.deleteSync()` at the start of `_download` runs once before the retry loop — each attempt resumes its own partial, never a stale artifact.

- `CI_ACTIONS_TOKEN` in `upgrade-check.yml` — only exposed via `secrets:` to a `schedule`/`workflow_dispatch`-triggered reusable workflow, never to fork PRs or in any `run:` block. The token is needed because `GITHUB_TOKEN` cannot push `.github/workflows/` changes.

- `record_use` dependency widened from `^0.6.0` to `>=0.6.0 <2.0.0` — accommodates the 1.0.0 release without a breaking version lock. Lockfile confirms resolution to 1.0.0.

- `yaml` promoted from transitive to direct main dependency — required by the new `readPdfManipulatorUserDefines()` for the web setup path, explicitly declared in pubspec.

- The `_compileWasm` function in `hook/build.dart` — the inline cargo build + bindgen_runner steps replaced with a call to the shared `compileWasmEngine()`. Temp dir naming changed from `pdf_manipulator_trim_wasm_` to `pdf_manipulator_wasm_`. Cleanup now keyed on `isDefaultBuild` (not just `featuresOverride != null`), correctly handling the `build: size`/`debug` case alongside feature overrides.

- The `compileEngineForTarget` Android `_AR` env in `engine_compiler.dart:1909-1914` — the `androidLinkerEnvFromNdkHome` function exports both `CARGO_TARGET_<triple>_LINKER` and `CARGO_TARGET_<triple>_AR` (pointing at the NDK's `llvm-ar`), matching `androidLinkerEnv` from the hook path. The old bash driver exported only the linker. Addressed from review findings on PR 185.

- `engine_build.dart` `cargoEnv` — `size` sets `CARGO_PROFILE_RELEASE_OPT_LEVEL=z`, `debug` sets `CARGO_PROFILE_RELEASE_DEBUG=2` + `CARGO_PROFILE_RELEASE_STRIP=none`. Both ride on the release profile so `panic=unwind` survives — native `catch_unwind` crash isolation keeps working. Correctly documented.

- `web_assets/` protection — `hook/build.dart:447-448`: only the default build (full features + speed) reads from or writes to `web_assets/`. A custom feature set or non-default build writes to a temp dir and cleans up after. This prevents a trimmed or size-build artifact from overwriting the committed default artifact.

- `bin/setup.dart` — the `--trim` flag removed. Config now read from the app's pubspec via the shared `PdfManipulatorConfig.parse` + `resolveKeepPlan`, using the same `readPdfManipulatorUserDefines` → `EngineBuild.parse` path native uses. Web and native config are now one block, one parser, one set of keys.

- All documentation: `README.md`, `CHANGELOG.md`, `CHANGELOG.pre.md`, `ARCHITECTURE.md`, `CAPABILITY_ROADMAP.md`, `UPDATING.md` — consistent with the new vocabulary and expanded to cover the `build:` axis and the size-table rows.

slopfairy (approved):
> [!TIP]
> Approved — clean promotion: unified compile driver, keep/build config surface, download resilience, and Rust version gating across 60 files

✅ **Checked**
- `.github/workflows/upgrade-check.yml` — passes `sweepActions: true` and `CI_ACTIONS_TOKEN` only to whuppi/ci's v2.4.0 reusable workflow, triggered by `workflow_call`/`workflow_dispatch`/`schedule` (no fork-reachable trigger). Every `uses:` ref in the repo is pinned to an exact SHA with a version comment.
- `lib/src/hook/pdf_config.dart` — the single parser validates all three keys (`keep`, `detector`, `build`), rejects unknown keys on web (native's known limitation is documented), rejects `detector` without `keep: auto`, rejects bool/map-style `keep` values with clear migration messages. Invalid configs are genuinely unrepresentable past this point.
- `lib/src/hook/engine_build.dart` — `speed`/`size`/`debug` all ride on cargo's `release` profile via env overrides, keeping `panic = "unwind"` intact so native crash isolation still works. Only `speed` has prebuilt binaries; `size` and `debug` compile from source.
- `lib/src/hook/engine_compiler.dart` — `ensureCargoVersion` gates the build against `RUST_VERSION` from `tool/versions.env`, failing with actionable `rustup` commands on a too-old cargo (fix for issue #183). The regex `cargo\s+(\d+\.\d+\.\d+)` parses version output correctly. `--no-default-features` added to the native compile path, symmetric with WASM — safe because the features string always seeds from build.json's `native` set which carries the crate defaults (`icc`, `legacy-crypto`).
- `lib/src/hook/resolver.dart` — download retries up to 4 attempts with exponential backoff + jitter and HTTP Range resume (206 handling). 404/410 exit early (asset genuinely absent); other failures retry. Stalled sockets time out at 60s so the retry loop advances rather than hanging.
- `hook/link.dart` — the RecordUse rebuild now carries `build.cargoEnv` so a `size`/`debug` consumer's trimmed rebuild matches the initial compile's profile (previously would silently revert to `speed`).
- `tool/compile.dart` — replaces `tool/compile_rust.sh` entirely. Calls the same `compileEngineForTarget`/`compileWasmEngine` core the consumer hook uses, so CI and consumer can't drift. Android NDK resolution via `androidLinkerEnvFromNdkHome` mirrors the old bash logic and sets both `_LINKER` and `_AR` env vars.
- `tool/analyze.sh` and `tool/shake_audit.sh` — feature strings now read from `build.json` via `json_get` instead of hand-duplicated strings or shelling out to a compile script. The shake audit's `merge_sizes` function merges (never overwrites) so partial runs don't drop columns.
- Keep/trim rename (`lib/src/trim/` → `lib/src/keep/`, `TrimConfig` → `KeepConfig`, etc.) is comprehensive — all imports, class names, method calls, and `KeepRecord.op` call sites in `lib/src/ops/` are renamed consistently. The `Grammar` for `keep: [...]` is now a flat list, not a nested map.
- CHANGELOG entries for v4.0.0 / v4.0.0-dev.0, README size table with `size` column, and architecture docs are accurate and complete.

PR #209, opened 2026-07-28
feat: scan-dirs, and make keep: auto actually scan your app

chaudharydeepanshu:
Three changes in the `keep: auto` lane. The first two are fixes to behaviour that was silently wrong; the third is the feature that closes the gap they exposed.

## 1. The scan was reading the wrong tree

A build hook is started with its working directory set to **the package's own root** — `hooks_runner`'s `build_runner.dart`: `final workingDirectory = input.packageRoot;`. So `Directory.current` was pdf_manipulator, not the app. The scan matched every capability where it is *implemented*, concluded the app used all of them, and kept the full binary — every time, while reporting success. `keep: auto` on native was effectively a no-op.

The app root is now derived from the hook's output directory (`<app>/.dart_tool/hooks_runner/…`), the one handle on the consuming app a hook is given. Web was never affected: `setup` runs from the user's own project, so `Directory.current` **is** the app there — `bin/setup.dart` carries a comment so the two lines are never "aligned".

## 2. The decision was cached forever

The hook registered none of the app's source as a build dependency, so once it had decided it never looked again — add a `sign()` call to an already-built app and the engine kept compiling without signature support.

Scanned roots and files are now registered. Both halves are load-bearing: the runner hashes a directory's **child names only** (`_hashDirectory`, `recursive: false`), so a directory catches a file appearing or disappearing but *cannot* see an edit — the per-file entries catch that.

Roots are also named source dirs only (`lib/`, `bin/`), never the bare app root. The staleness check takes a directory's `lastModified` **recursively**, so a dependency covering `.dart_tool/` — which the build itself rewrites — would force a full engine recompile on every single build.

## 3. New: `scan-dirs`

```yaml
hooks:
  user_defines:
    pdf_manipulator:
      keep: auto
      scan-dirs: [tools, packages/shared/lib]
```

Code outside `lib/` and `bin/` was invisible to the scan, so a capability you really call could be trimmed away. Now you point at it.

Validated per the config parser's own rule that a stranded option is never silently ignored — it needs a scan to widen, so it is rejected without `keep: auto`, and rejected with `detector: record-use` (which reads post-AOT recordings and never opens a source file). Allowed with `compare`, which does trim via the scan. Paths are app-relative; absolute paths are rejected because a committed pubspec is shared and would be right on exactly one machine. **A directory that isn't there falls closed to the full binary** — a typo must not quietly narrow the scan into a confident wrong trim.

## Version

Both changelog lanes move to **4.1.0** / **4.1.0-dev.0**: the untagged `4.0.1` entry now carries an added capability, and the rules say *Added / Changed capability → MINOR* with *exactly ONE new untagged version at the top*. `release.sh --check-versions` passes.

## Verification

- `dart analyze --fatal-infos` clean, `dart format` clean, **481/481 tests pass**.
- 8 new tests. The run log is the proof: widened → `keeping {extract (lib/app.dart:6), getSignatures (tools/report.dart:4)}`; typo'd dir → `1 path(s) could not be read — keeping the FULL binary (fail closed)`.
- Two guard tests pin the wiring itself (the hook must not read `Directory.current`, and must register `appDependencies`) — the defect lived on a line no unit test could reach.

## Also included

`docs/ARCHITECTURE.md` gains a section documenting that editor mutations stage and materialise at `save_to_bytes()`, while the destructive ops (`applyRedactions`, `eraseRegions`) do their work at call time and stage the *result*. Unrelated to the keep work; it is a doc-only addition.


slopfairy:
posted my review below 👇

lib/src/keep/detector.dart
```diff
@@ -53,14 +65,29 @@ class DetectorResult {
...
+/// capabilities: its `lib/` and `bin/`, plus any [extraDirs].
+///
+/// Only those directories are read, and only the app's own source — a
+/// capability reached from a package the app depends on is invisible. So
+/// `resolved` means "nothing was unreadable", never "every capability the
+/// app uses was seen": an app whose source lives elsewhere scans clean and
+/// under-keeps. Callers that cannot widen the scan should say so in their
+/// own docs and point users at an explicit keep-set.
 DetectorResult detectCapabilities(
   String appRoot, {
```
slopfairy, line 87: The scan roots (`scannedRoots`) only include top-level directories: `lib/`, `bin/`, and any `scan-dirs` entries. The runner's `_hashDirectory` with `recursive: false` hashes child *names* only, so adding a new `.dart` file inside a subdirectory (e.g. `lib/models/`) does not change the hash of the registered root — the hook never re-runs and the new capability calls are invisible.

Only new files at the top level of `lib/` or `bin/` invalidate the cached decision. A new file in a subdirectory is an edit to the app that goes undetected — the same class of bug as the old code's "cached forever" defect, just narrower.

Consider registering every directory that contains at least one `.dart` file (not just the top-level roots) as a `Uri.directory` dependency. The number of extra dependencies is bounded by the number of scanned files, and none of them are build-volatile (the `.` / `build` skip already keeps those out).

slopfairy (dismissed):
> [!TIP]
> Approved — fixes two silent correctness bugs in `keep: auto` and adds `scan-dirs`; sound design, thorough tests

1. `lib/src/keep/detector.dart:78` — the `scannedRoots` list only includes top-level directories (`lib/`, `bin/`, scan-dirs entries). The runner hashes directory child *names* only (`recursive: false`), so adding a new `.dart` file inside a subdirectory (e.g. `lib/models/`) does not invalidate the cached decision — only new files at the top level of a registered root do. Consider registering every directory that contains at least one `.dart` file as a `Uri.directory` dependency, not just the top-level roots. The `.` / `build` skip already keeps build-volatile paths out, so every such directory is safe to register.

✅ **Checked**
- `appRootFromHookOutput` correctly derives the app root from the last `.dart_tool` segment in the hook's output directory path, and returns null when the segment is absent or at position 0 (filesystem root — never a plausible app)
- The fix 1 diagnosis is verified: the build runner sets `workingDirectory = input.packageRoot`, so `Directory.current` inside the hook was pdf_manipulator, and the scan matched every capability's implementation
- Web path (`bin/setup.dart`) correctly keeps `Directory.current` — the user runs `setup` from their own project, so the working directory IS the app. The comment guards against future "alignment" refactors
- Config validation is thorough: `scan-dirs` is rejected without `keep: auto`, rejected with `detector: record-use` (never reads source), and allowed with `compare` (uses the scan). Absolute paths are rejected with a clear message about shared pubspecs
- A missing scan-dirs entry falls closed to the full binary rather than quietly narrowing the scan into a confident wrong trim
- The directory walk skips `.`-prefixed dirs and `build/` — this prevents `.dart_tool/` (which the build rewrites) from ever becoming a hook dependency and forcing a full engine recompile on every build
- Both halves of dependency registration are load-bearing: `Uri.directory(root)` catches files appearing/disappearing at the root level, `Uri.file(file)` catches edits to existing files (directory mtime doesn't change on edit)
- Guard tests pin the wiring itself: `hook/build.dart` must not contain `Directory.current` and must contain `appDependencies` after stripping comments. The defect lived on a line unit tests couldn't reach, and now it can't regress
- `KeepPlan.appDependencies` is correctly empty for modes that read no app source (manual keep-set, `keep: all`, `record-use`)
- Version bump to 4.1.0 / 4.1.0-dev.0 is correct per the project's MINOR rule (added capability)
- `docs/ARCHITECTURE.md` editor mutation docs are clean and well-placed — doc-only addition, no code impact

slopfairy:
posted my review below 👇

chaudharydeepanshu:
Verified and fixed in 6435ba0 — you were right, and it was live.

`findOutdatedDependency` (the between-builds check, not just the update path) calls `_hashDirectory`, which is `listSync(recursive: false)` over child **names**. So registering `lib/` alone caught a file added directly in `lib/`, but a file added in `lib/models/` left `lib/`'s child names untouched — and the new file wasn't in `scannedFiles` either, since it didn't exist when the scan ran. The stale decision survived. `scannedRoots`' own doc already claimed "the directories walked", so the doc and the code disagreed.

One refinement to the suggestion: *"every directory that contains at least one .dart file"* isn't sufficient. If `lib/assets/` holds no Dart today and someone adds `lib/assets/thing.dart`, `lib/`'s child names still don't change — so it has to be **every directory the walk could read**, empty ones included. That's what landed; `scannedRoots` is now `scannedDirs`.

Agreed on the safety point: the `.`/`build` skip already keeps build-volatile trees out, so everything the walk reaches is safe to register.

Pinned by a test that fails against the old behaviour — reverted `scannedDirs` to `roots` locally to confirm: `Expected: contains '…/lib/models/'`. It asserts `lib/`, `lib/models/`, and the Dart-less `lib/assets/`.

`dart analyze --fatal-infos` clean, format clean, **482/482**.

slopfairy (dismissed):
> [!TIP]
> Approved — scan-dirs works, the two silent-behaviour bugs are fixed, and the prior review's subdirectory-dependency gap is closed

✅ **Checked**
- `appRootFromHookOutput`: derives app from hook output path, null-safe, correct `lastIndexOf` for nested `.dart_tool`
- `detectCapabilities`: every directory walked is tracked in `scannedDirs`; `.` and `build` dirs skipped; missing extra-dirs fall closed
- `resolveKeepPlan`: `appDependencies` includes pubspec, every scanned dir, and every scanned file — directory hashing catches new files, file hashing catches edits
- `hook/build.dart`: wired to `appRootFromHookOutput` (not `Directory.current`), `output.dependencies.addAll(plan.appDependencies)` called
- `bin/setup.dart`: comment guards the different working-directory contract; `scanDirs` passed through
- `pdf_config.dart`: `scan-dirs` validation complete — rejects absent `keep: auto`, rejects `detector: record-use`, allows `compare`, rejects absolute paths, non-list, empty entries
- Config tests: 8 cases covering parse, rejection, and cross-axis coupling
- Integration tests: 10 cases on the full seam from hook output to keep plan, including subdirectory registration, typo fall-closed, scan-dirs widening
- Guard tests: hook source inspected for `appRootFromHookOutput` presence and `Directory.current` absence, `appDependencies` presence
- CHANGELOG / README: clear migration note about `keep: auto` now actually trimming

slopfairy:
posted my review below 👇

slopfairy (dismissed):
> [!TIP]
> Approved — the three changes are correct, the fix from the previous round (scannedDirs registration) landed with tests, and the new tests cover the seams the old code missed.

1. `lib/src/keep/detector.dart:140` — the `.` prefix skip (`name.startsWith('.')`) is broad enough to skip any user directory under `lib/` that happens to start with a dot, like `lib/.generated/`. In practice nobody puts Dart source there and `scan-dirs` is the escape hatch, so this is the right default — just noting the blast radius for future readers.

✅ **Checked**
- `appRootFromDartTool` correctly uses `lastIndexOf('.dart_tool')` with the `< 1` guard that handles both "absent" and "at filesystem root"
- Config validation for `scan-dirs`: rejected without `keep: auto`, rejected with `detector: record-use`, allowed with `compare`, absolute paths rejected, non-list/empty/non-string entries rejected
- `detectCapabilities` now records every walked directory (not just roots) and every `.dart` file, skips `.` and `build` subtrees, and records missing extra dirs as unresolved
- `keep_plan.dart`: `appRootCandidate` is now nullable, `scanDirs` added, `appDependencies` computed from pubspec + all scanned dirs + all scanned files — empty for manual/explicit keep modes
- `hook/build.dart`: uses `appRootFromDartTool(input.outputDirectory)` instead of `Directory.current`, registers `plan.appDependencies`
- `bin/setup.dart`: derives app root from `Isolate.packageConfig` via `appRootFromDartTool`, guard exits with clear error on null
- Tests cover: both lane anchors resolve correctly, trim works end-to-end from hook output path, no volatile paths leak into dependencies, every subdirectory registered (not just roots), scan-dirs widens and missing dir fails closed, explicit keep-set has empty dependencies, hook wiring pinned with source-level assertions
- Changelog entries (both stable and pre-release) at 4.1.0 / 4.1.0-dev.0 are accurate and follow the repo's versioning rules

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — three tightly-scoped changes in the `keep: auto` lane, all correct

✅ **Checked**

- `appRootFromDartTool` — uses last `.dart_tool/` segment (innermost wins for nested projects), falls through to null when absent; both callers treat null as "build everything." The same function serves the hook and `setup`, so web and native cannot disagree about where the app is.
- Hook dependency registration — every walked directory + every scanned file + `pubspec.yaml`. Directories catch file appearance/disappearance (runner hashes child names, `recursive: false`); files catch edits. Together they close the staleness gap the old code had.
- Named source dirs only (`lib/`, `bin/`), never the bare app root. A dependency covering `.dart_tool/` — which the build itself rewrites — would force a full engine recompile on every build under `keep: auto`.
- `_toolDirs` (`{.dart_tool, .git, .fvm}`) — skip by name, not prefix. A user's `lib/.generated/` is scanned; a generated tree is not. The `build/` skip is pinned to the app root only; `lib/build/` is app source.
- `scan-dirs` config validation — rejected without `keep: auto`, rejected with `detector: record-use`, allowed with `detector: compare`. Same "stranded option is never silently ignored" rule the parser already enforces for `detector`.
- Absolute paths in `scan-dirs` are rejected (`_parseScanDirs` checks Unix `/` prefix and Windows `C:\`). A committed pubspec is shared — a machine-specific path breaks CI.
- Missing `scan-dir` falls closed to the full binary — the extra dir check in `detectCapabilities` records it as unresolved, `resolveKeepPlan` sees `!result.resolved` and builds everything. A typo cannot quietly narrow the scan.
- Test coverage — hook_app_root_test.dart is 330 lines covering the seam no existing test reached: the path from hook output directory to keep decision. Pins both `appRootFromDartTool` correctness AND the wiring guard (no `Directory.current` in the hook). The config parser tests cover every `scan-dirs` rejection path.
- Link hook unchanged — correct, it never reads app source.
- CHANGELOG lanes consistent, version follows semver rules.

PR #210, opened 2026-07-28
release: promote dev → prod (v4.1.0)

chaudharydeepanshu:
Promotes `dev` to `prod`, cutting stable **v4.1.0**. `CHANGELOG.md`'s top heading moves `4.0.0` → `4.1.0`, which triggers the release pipeline on merge.

> [!IMPORTANT]
> **Merge with a merge commit — not squash, not rebase.** Per `docs/UPDATING.md` S6, the promotion PR must be a merge commit; a squash here diverges prod from dev and needs the force-sync recovery procedure.

## What ships

One commit — [#209](https://github.com/whuppi/pdf_manipulator/pull/209) — the `keep: auto` lane.

**Two fixes to behaviour that was silently wrong:**

- **`keep: auto` kept everything on native, every time.** A build hook is started with its working directory set to the *package's own* root (`hooks_runner`: `workingDirectory = input.packageRoot`), so reading `Directory.current` scanned pdf_manipulator — where every capability is implemented — and concluded the app used all of them, while reporting success. Verified end-to-end after the fix: an app calling only `extract` now compiles `["default", "extract", "icc", "icc-qcms", "legacy-crypto", "native-bridge"]` — **8.9 MB against 20 MB** for the full build.
- **The decision was cached forever.** No app source was registered as a build dependency, so once decided it never looked again. Now every scanned directory *and* file is registered — both halves matter, because the runner hashes a directory's immediate child names only, so directories catch a file appearing and files catch an edit.

**One feature:**

- **`scan-dirs`** — extra directories for the scan, on top of `lib/` and `bin/`, for code that lives in `tools/` or a sibling package. Validated like `detector` is: rejected without `keep: auto`, rejected with `detector: record-use`, app-relative paths only, and a directory that isn't there falls closed to the full binary.

**Plus, from review on #209:**

- Web `setup` now locates the app the same guaranteed way native does (a path inside the app's own `.dart_tool/`), instead of `Directory.current` — it had silently ignored your config when run from a subfolder and shipped the full web engine.
- Tool trees are skipped **by name** (`.dart_tool`, `.git`, `.fvm`, and `build/` only at the app root) rather than by `.`-prefix, so a dot-directory you created — `lib/.generated/` — is your source and gets scanned.

## Upgrade note for users

`keep: auto` on native starts genuinely trimming, so a build that had been silently shipping every capability gets a smaller one. The changelog leads with this: verify the app after upgrading, and state anything missing with `keep: [render, …]`.

## Verification

`dart analyze --fatal-infos` clean, `dart format` clean, **483/483 tests**. Exercised live on a throwaway consumer app for both lanes — native trimmed to `extract`, and web run *from a subfolder* producing `keep: features [wasm,extract]`.


slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — release promotion of the previously-reviewed `keep: auto` lane from PR #209, cleanly staged for stable v4.1.0.

✅ **Checked**
- `CHANGELOG.md` / `CHANGELOG.pre.md` — 4.1.0 entries match the shipped content from #209; 4.0.0 heading stays in place below the new entry, so the release pipeline's heading-change trigger fires correctly on merge.
- `README.md` — `scan-dirs` example is accurate and the limitation note ("no scan can see a call made inside a package you depend on") is honest.
- `hook/build.dart` — app root is now derived from `input.outputDirectory` via `appRootFromDartTool`, not `Directory.current`, and `output.dependencies` registers `plan.appDependencies`.
- `bin/setup.dart` — web lane also derives the app root from `Isolate.packageConfig` via the same `appRootFromDartTool` function, fixing the subfolder-config-silence bug, and passes `scanDirs` through.
- `lib/src/hook/app_root.dart` — shared derivation uses `lastIndexOf('.dart_tool')` to handle nested tool directories correctly; returns null cleanly when absent.
- `lib/src/hook/keep_plan.dart` — `appRootCandidate` is now nullable, `scanDirs` parameter added, dependencies are collected from the detector result, and both `null` and missing-pubspec fall closed.
- `lib/src/hook/pdf_config.dart` — `scan-dirs` validation: rejects without `keep: auto`, rejects with `detector: record-use`, rejects absolute paths, rejects non-list and empty entries, added to `knownKeys`.
- `lib/src/keep/detector.dart` — tool trees skipped by name (`.dart_tool`, `.git`, `.fvm`; `build/` only at app root), not by `.`-prefix; every walked directory and scanned file recorded; missing `extraDirs` and empty `roots` both fall closed; no bare app root scan (avoids `.dart_tool/` churn).
- Tests — `hook_app_root_test.dart` (330 lines) covers the full seam: app root derivation from both lane anchors, actual trimming from a real temporary app, no volatile dependencies, directory and file registration, dot-directory handling, `scan-dirs` widening, missing `scan-dir` fall-closed, explicit keep-set emptiness, and source-level guards on `hook/build.dart`. `pdf_config_test.dart` covers all `scan-dirs` parse cases. Existing `keep_plan_test.dart` updated with the new parameter.
- The merge commit instruction in the PR body ("merge commit — not squash, not rebase") matches the documented promotion procedure in `docs/UPDATING.md`.

slopfairy:
posted my review below 👇

slopfairy (approved):
> [!TIP]
> Approved — clean release promotion, substantive changes already reviewed in #209

Note: this PR was merged while the review was in progress.

✅ **Checked**
- CHANGELOG.md (4.1.0) and CHANGELOG.pre.md (4.1.0-dev.0) entries are consistent — same content, correct heading levels
- `appRootFromDartTool` in `lib/src/hook/app_root.dart` uses `lastIndexOf('.dart_tool')` — handles the nested case correctly, and null / no-marker / root-marker all return null with callers falling closed to the full binary
- Tool tree skipping in `lib/src/keep/detector.dart` is by name (`.dart_tool`, `.git`, `.fvm` anywhere; `build/` only at the app root), not by `.`-prefix — a user's `lib/.generated/` is correctly treated as app source
- `scan-dirs` validation in `lib/src/hook/pdf_config.dart`: rejected with non-`auto` keep, rejected with `detector: record-use`, absolute paths rejected, empty/non-string entries rejected, non-list rejected
- Missing `scan-dir` falls closed to the full binary (not a silent narrow scan) — enforced in both `detector.dart` (records the unresolved path) and `keep_plan.dart` (checks `result.resolved`)
- Build hook in `hook/build.dart` registers `plan.appDependencies` via `output.dependencies.addAll(...)` — every scanned file and directory becomes a build dependency, so the keep decision re-evaluates when app source changes
- Web `setup` in `bin/setup.dart` uses `Isolate.packageConfig` + `appRootFromDartTool` instead of `Directory.current` — same anchor derivation as the build hook, so web and native cannot disagree about which directory the app is
- README.md docs for `scan-dirs` match the implementation
- `docs/ARCHITECTURE.md` editor mutations section is accurate documentation
- Tests cover both lane anchor resolution, driven trim correctness, volatile path exclusion, dependency registration (directories + files), dot-directory scanning, scan-dirs widening, typo fall-closed, explicit keep-set having no dependencies, and the wiring assertions that `hook/build.dart` derives the app root rather than reading `Directory.current`

PR #218, opened 2026-08-29
fix: checked checkboxes and selected radios flatten as set (#215)

### LINKED ISSUE #215: setFormFieldValue cannot check checkboxes and flattenForms bakes the unchecked appearance
## Summary

  With `pdf_manipulator 4.1.0`, setting a checkbox using
  `setFormFieldValue()` and then calling `flattenForms()` produces a flattened
  PDF where the checkbox is still visibly unchecked.

  This also reproduces with a checkbox created by `pdf_manipulator` itself, so it
  does not appear to depend on a particular third-party PDF producer.

  ## Expected result

  The flattened PDF contains a visibly checked checkbox.

  ## Actual result

  The flattened PDF contains only the empty checkbox border.

  As a control, creating the checkbox with `checked: true` and then flattening it
  does preserve the checkmark. The failure is specific to changing the checkbox
  through `setFormFieldValue()`.

  ## Structural evidence

  The checkbox filled using:

  ```dart
  await editor.setFormFieldValue('agree', 'Yes');
  ```

  produces the same 56-byte unchecked appearance stream as an untouched
  checkbox:

  ```pdf
  1 1 1 rg
  0 0 20 20 re f
  0 0 0 RG
  1 w
  0.5 0.5 19 19 re S
  ```

  A checkbox created with `checked: true` produces a 92-byte stream containing
  the checkmark path:

  ```pdf
  4 10 m 8 4 l 16 16 l S
  ```

  ## Suspected cause

  The public Dart API only accepts a string:

  ```dart
  PdfTask<void> setFormFieldValue(String fieldName, String value)
  ```

  The native dispatch then unconditionally creates a text value:

  ```rust
  editor.set_form_field_value(
      name,
      FormFieldValue::Text(value.to_string()),
  )
  ```

  The engine's checkbox implementation expects:

  ```rust
  FormFieldValue::Boolean(true)
  ```

  During flattening, the `"Yes"` text updates `annotation.field_value`, but the
  existing `WidgetFieldType::Checkbox { checked: false }` remains unchanged.
  Appearance generation reads that boolean and therefore generates the unchecked
  appearance.

  ## Possible fix

  Expose typed checkbox values through the Dart bridge, for example:

  ```dart
  editor.setCheckboxFieldValue('agree', true);
  ```

  or use a typed form-field value API supporting both text and boolean values.

  The bridge should pass `FormFieldValue::Boolean` for checkboxes, update the
  widget's checked state before regenerating its appearance, and serialize `/V`
  and `/AS` using the checkbox's actual non-`/Off` appearance-state name. The
  latter matters for PDFs whose checked export value is not `/Yes`.

### Steps to reproduce

  ```dart
  import 'dart:io';
  import 'dart:typed_data';

  import 'package:flutter_test/flutter_test.dart';
  import 'package:pdf_manipulator/pdf_manipulator.dart';

  void main() {
    test('filled checkbox remains visible after flattening', () async {
      final pdf = Pdf();
      addTearDown(pdf.dispose);

      final template = await buildCheckbox(pdf);

      final editor = await pdf.edit(MemorySource(template));
      try {
        await editor.setFormFieldValue('agree', 'Yes');
        await editor.flattenForms();

        final sink = MemorySink();
        await editor.save(sink);

        await File('checkbox-filled-flattened.pdf').writeAsBytes(
          sink.takeBytes(),
        );
      } finally {
        await editor.dispose();
      }
    });
  }

  Future<Uint8List> buildCheckbox(Pdf pdf) async {
    final builder = await pdf.build();
    try {
      final page = await builder.addA4Page();
      await page.checkbox(
        'agree',
        const PdfRect(x: 72, y: 660, width: 20, height: 20),
      );
      await page.done();

      final sink = MemorySink();
      await builder.save(sink);
      return sink.takeBytes();
    } finally {
      await builder.dispose();
    }
  }
  ```

### Target (which platform?)

- [x] Android
- [ ] iOS
- [ ] macOS
- [x] Linux
- [ ] Windows
- [x] Web

### Package version

4.1.0

### Flutter version

 3.44.4

### Did this work before?

None

chaudharydeepanshu:
Closes #215, reported by [@DarkWingMcQuack](https://github.com/DarkWingMcQuack).

Setting a checkbox with `setFormFieldValue` and flattening produced a visibly unchecked box. The report identified one cause; tracing it end to end found five, and the suggested fix would not have worked on its own.

## A button's value is a name, not text

A checkbox or radio is on when its value names one of the appearance states in the widget's `/AP /N` dictionary. `/V` names the chosen state, and each widget's `/AS` says which state it is currently showing. Every point that consumed the value treated it as text instead.

**1. The Dart API could only send a string.** `setFormFieldValue(String, String)` was the only door, so the value arrived as `FormFieldValue::Text`.

**2. The flattener ignored button state.** This is the one the report's suggested fix would have missed. `modified_widget_text` matches only `Text | Choice`, so even with `FormFieldValue::Boolean` on the wire it returns `None`, the flattener falls through to `extract_widget_appearance`, reads the document's stale `/AS` (still `/Off`), and generates from `Checkbox { checked: false }`. The empty box gets baked in either way.

**3. `/V` and `/AS` were written as text strings.** `Object::from(&FormFieldValue)` emits a text string for the string lane, so the file got `/V (Yes)` — which names no appearance state. The value was lost on reopen too, not just when flattening.

**4. The on-state was hardcoded `/Yes`.** `/Yes` is a convention, not a rule. A box whose state is `/On` or `/1` never matched.

**5. Radio and Pushbutton `/Ff` bits were swapped.** ISO 32000-1 Table 227 numbers bit positions from 1, so Radio is position 16 (`0x8000`) and Pushbutton is 17 (`0x10000`). `annotations.rs` read them the other way round, so every spec-conformant radio group parsed as a push button — which has no `/AS` to track. `extractors/forms.rs` already had the correct constants; the fix uses that module rather than re-spelling the literals, so the two cannot drift apart again.

Radio had a sixth problem on top: `/AS` was written to the parent field, but a radio group's states live on its `/Kids`, so a selection never reached a widget.

## What changed

A button's value is now read as a state name and resolved against the states that widget actually offers. `/V` and `/AS` are written as names, `/AS` reaches each kid, and flattening draws the state that was set — preferring the widget's own artwork over a regenerated mark, and regenerating only when there is no state dictionary to pick from.

Resolution is deliberate about the two ambiguous cases. An exact match against the offered states always wins first, because a Yes/No radio group really does have a state named `/No` — the word can only be read as "off" once the widget has said it offers no such state. After that a word that plainly means on or off is taken at its word, and any other unrecognised name selects nothing rather than guessing.

New API for the caller who knows a box should be on but not what its on-state is called:

```dart
await editor.setCheckboxFieldValue('agree', true);
```

`setFormFieldValue` still takes the name directly when you do know it, and now reads it as a name.

**This is not Dart-only.** The defects were in the shared engine, so the Python, Swift, Go and JS bindings are fixed too. The JNI binding is the clearest case: `pdf_oxide_jni/src/editor.rs:125` already had a correct boolean checkbox setter, but it was passing a good value into a path where nothing downstream honoured it. It works now without a line of binding code changing.

## Tests

Written first and watched fail: the control passed while all four bug cases reported `0.0` dark pixels. **13 Rust + 10 Dart**, all passing now.

Proof is semantic, per this repo's convention — the flattened page is rasterised and ink measured in each widget's band, against a hand-authored fixture whose on states fill solid black and whose off states do not. Hand-authored because no foreign writer produces the preconditions: dart-pdf always names a checkbox's on state `/Yes` and emits no radio group at all.

Covered: the string lane and the boolean lane, `/Yes` and `/On` boxes, radio selection with sibling exclusion, off words (`Off`, `off`, `No`, `false`, `0`), an unrecognised name selecting nothing, an offered state named `/No` still being selectable, `/V` and `/AS` written as names, and a value set before save surviving a reopen-then-flatten. The engine repo carries its own guard in `tests/form_fill_button_state.rs` so the fix cannot regress from that side either.

The coverage hole that let this ship: `test_form_extraction.rs:495` already called `set_form_field_value(… Boolean(true))`, but only asserted the stored `/V` after a save. It never flattened.

**Tests:** yes — 13 Rust + 10 Dart, all failing before and passing after. Not target-specific; they pass on native and on all three web modes.

**Breaking:** no. The previous behaviour on a button field did nothing useful, so nothing could depend on it.

## Verification

`make analyze` (including the Rust warning gate on patched lines, native + wasm), the engine suite (337 binaries), native ops (242), web ops on all three modes, `test-unit` (178), `test-example` on VM + macOS, `verify-tarball`, and the wire-parity test.

Two pre-existing failures on `dev` are unrelated and were confirmed by re-running with these changes stashed: two web-atomics dispose stress tests (5000-iteration timing), and `test-example-web`, which needs chromedriver running locally. `make test-guards` is also red on `dev`; #217 fixes that separately.

Engine changes are in `vendor/pdf_oxide` (`3425288c` on `pdf_manipulator/0.3.73-patches`), each marked with the `pdf_manipulator patch` boundaries per `docs/UPDATING.md` S2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,13 @@ CONTENT RULES (never change)
 
 <!-- Add new versions below, newest first. -->
 
+## 4.2.0
+
+- Engine updated — web: re-run `flutter pub run pdf_manipulator:setup --force web` (native updates itself)
+- Added `setCheckboxFieldValue(fieldName, checked)` for a checkbox whose on-state name you do not know. A checkbox is on when its value names one of the states in its `/AP /N` dictionary, and that name is a convention rather than a rule — `/Yes` is the common one, `/On` and `/1` are not rare. Passing `true` picks whichever name that widget offers. `setFormFieldValue` still takes the name directly when you do know it.
+- Fixed a checkbox flattening as an empty box after it was checked. `setFormFieldValue` passed every value through as text, so a checkbox kept the unchecked state it was parsed with and `flattenForms` baked that empty appearance into the page. `/V` and `/AS` were written as text strings too, which name no appearance state, so the value was also lost when the file was opened again. A button's value is now read as a state name, resolved against the states that widget actually offers, and written as a name; flattening draws the state that was set, using the widget's own artwork where the document supplies it ([#215](https://github.com/whuppi/pdf_manipulator/issues/215) reported by [@DarkWingMcQuack](https://github.com/DarkWingMcQuack))
+- Fixed radio groups being impossible to select. `/AS` was written to the parent field, but a radio group's appearance states live on its kid widgets, so nothing changed on the page. The two button flags were also swapped when reading a widget — ISO 32000-1 puts Radio at bit position 16 and Pushbutton at 17 — so a spec-conformant radio group was read as a push button, which has no state to track. Setting a group's value now turns on the kid that offers that state and clears its siblings
+
 ## 4.1.0
 
 - Added `scan-dirs`: extra directories for `keep: auto` to scan, on top of your app's `lib/` and `bin/`. Code living somewhere else — a `tools/` folder, a sibling package in a monorepo — was invisible to the scan, so a capability you really call could be trimmed away. Now you point at it: `scan-dirs: [tools, packages/shared/lib]`, paths relative to your app. It only applies where there is a scan to widen (`keep: auto`, and not `detector: record-use`), and a directory that isn't there keeps the full binary rather than quietly scanning less than you asked for.
--- a/CHANGELOG.pre.md
+++ b/CHANGELOG.pre.md
@@ -81,6 +81,13 @@ CONTENT RULES (never change)
 
 <!-- Add new versions below, newest first. -->
 
+## 4.2.0-dev.0
+
+- Engine updated — web: re-run `flutter pub run pdf_manipulator:setup --force web` (native updates itself)
+- Added `setCheckboxFieldValue(fieldName, checked)` for a checkbox whose on-state name you do not know. A checkbox is on when its value names one of the states in its `/AP /N` dictionary, and that name is a convention rather than a rule — `/Yes` is the common one, `/On` and `/1` are not rare. Passing `true` picks whichever name that widget offers. `setFormFieldValue` still takes the name directly when you do know it.
+- Fixed a checkbox flattening as an empty box after it was checked. `setFormFieldValue` passed every value through as text, so a checkbox kept the unchecked state it was parsed with and `flattenForms` baked that empty appearance into the page. `/V` and `/AS` were written as text strings too, which name no appearance state, so the value was also lost when the file was opened again. A button's value is now read as a state name, resolved against the states that widget actually offers, and written as a name; flattening draws the state that was set, using the widget's own artwork where the document supplies it ([#215](https://github.com/whuppi/pdf_manipulator/issues/215) reported by [@DarkWingMcQuack](https://github.com/DarkWingMcQuack))
+- Fixed radio groups being impossible to select. `/AS` was written to the parent field, but a radio group's appearance states live on its kid widgets, so nothing changed on the page. The two button flags were also swapped when reading a widget — ISO 32000-1 puts Radio at bit position 16 and Pushbutton at 17 — so a spec-conformant radio group was read as a push button, which has no state to track. Setting a group's value now turns on the kid that offers that state and clears its siblings
+
 ## 4.1.0-dev.0
 
 - Added `scan-dirs`: extra directories for `keep: auto` to scan, on top of your app's `lib/` and `bin/`. Code living somewhere else — a `tools/` folder, a sibling package in a monorepo — was invisible to the scan, so a capability you really call could be trimmed away. Now you point at it: `scan-dirs: [tools, packages/shared/lib]`, paths relative to your app. It only applies where there is a scan to widen (`keep: auto`, and not `detector: record-use`), and a directory that isn't there keeps the full binary rather than quietly scanning less than you asked for.
--- a/docs/UPDATING.md
+++ b/docs/UPDATING.md
@@ -221,6 +221,34 @@ make check
 **Verify zero warnings:** `make analyze` checks Rust warnings in our
 patched lines automatically (see [S7](#s7--verify-our-code-warnings)).
 
+**Never add a test inside a fork.** A patch proves itself from this repo,
+in `test/ops/` — with a fixture in `test/fixtures/handwritten.dart` when
+the engine bug needs a PDF no foreign writer emits. A test file in
+`vendor/` is extra surface every S1 rebase has to carry onto the next base
+tag, for a proof the Dart battery already makes at the level a consumer
+actually experiences (rendered pixels, per ARCHITECTURE.md's testing
+section). The fork carries **patches only** — `src/`, and the build files a
+patch needs.
+
+Prior art: #161 was the same shape as #215 — an engine-side forms bug —
+and landed as a hand-authored fixture plus battery cases here, with the
+submodule bumped and no Rust test.
+
+If a proof genuinely cannot be made from Dart, that is a signal the fix
+belongs upstream: open it as a PR on the upstream repo, where its test
+lives with it (see `universal/external-contributions.md`).
+
+**Watch for a detached submodule.** `git submodule update` leaves
+`vendor/pdf_oxide` on a detached HEAD, so a commit made after it lands on
+no branch and `git push` reports "Everything up-to-date" while pushing
+nothing. Check before committing, and re-point the branch if it happened:
+
+```sh
+cd vendor/pdf_oxide
+git branch --show-current           # empty = detached
+git branch -f <branch-name> HEAD && git checkout <branch-name>
+```
+
 **Commit AND push the submodule, or CI will fail:**
 
 ```sh
--- a/lib/src/bridge/pdf_bridge.dart
+++ b/lib/src/bridge/pdf_bridge.dart
@@ -253,6 +253,9 @@ abstract class BridgeEditorHandle {
   /// Sets a form field's value by [fieldName].
   PdfTask<void> setFormFieldValue(String fieldName, String value);
 
+  /// Checks or clears the checkbox field identified by [fieldName].
+  PdfTask<void> setCheckboxFieldValue(String fieldName, bool checked);
+
   /// Crops margins from all pages.
   PdfTask<void> cropMargins({
     double left = 0,
--- a/lib/src/bridge/shared_bridge_editor.dart
+++ b/lib/src/bridge/shared_bridge_editor.dart
@@ -217,6 +217,12 @@ class _SharedEditorHandle extends BridgeEditorHandle {
   PdfTask<void> setFormFieldValue(String fieldName, String value) =>
       _mutate('setFormFieldValue', {'fieldName': fieldName, 'value': value});
   @override
+  PdfTask<void> setCheckboxFieldValue(String fieldName, bool checked) =>
+      _mutate('setCheckboxFieldValue', {
+        'fieldName': fieldName,
+        'checked': checked,
+      });
+  @override
   PdfTask<void> cropMargins({
     double left = 0,
     double right = 0,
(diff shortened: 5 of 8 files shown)


vendor/pdf_oxide
```diff
@@ -1 +1 @@
-Subproject commit a634d9f5b15cd0acb320fe926f33d8f29ef1a95b
+Subproject commit 3425288c8247571e3b61a56a6e38b69f411a8f95
```
slopfairy: The engine commit this pins (`3425288c`) does not compile. In `flush_form_fields_to_modified_objects` (`src/editor/document_editor.rs`), `new_value` is an owned `FormFieldValue`, but two call sites pass it by value where `&FormFieldValue` is required:

- `button_on_state(new_value, &available)` — `button_on_state` is declared `fn button_on_state(value: &FormFieldValue, available: &[String]) -> Option<String>`.
- `Object::from(new_value)` — only `From<&FormFieldValue> for Object` exists in `form_fields.rs`, not `From<FormFieldValue>`.

Fix both call sites to `&new_value`. (Correction to my earlier note: `Some(val) => val.clone()` is fine — it returns the owned value, not `&FormFieldValue`; the errors are only these two call sites.)

`Rust tests` will fail on this until the submodule is bumped to a compiling commit.

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the pinned engine commit does not compile

1. `vendor/pdf_oxide` (submodule `3425288c`) — `flush_form_fields_to_modified_objects` in `src/editor/document_editor.rs` clones the wrong thing. `val` binds as `&FormFieldValue`, so `val.clone()` returns `&FormFieldValue` (the blanket `Clone for &T` impl), which mismatches the `FormFieldValue` annotation and the `Vec<(u32, u16, FormFieldValue, bool)>` element type. Two call sites then pass the owned value where a reference is required:

```rust
// clone the value, not the reference
Some(val) => (*val).clone(),

// borrow at the call sites
let target = Self::button_on_state(&new_value, &available);
// ...
new_dict.insert("V".to_string(), Object::from(&new_value));
```

The last one matters because only `From<&FormFieldValue> for Object` exists (`form_fields.rs`), not `From<FormFieldValue>`. The `Rust tests` CI job is still in progress, so this has not been caught yet; the claimed `make analyze` and 337-binary run cannot reflect this exact commit.

✅ **Checked**
- Dart API surface — `setCheckboxFieldValue` is wired consistently across `pdf_bridge.dart`, `shared_bridge_editor.dart`, `pdf_editor.dart`, `bridge_api.rs`, and `dispatch.rs`.
- State resolution (`button_on_state`) — exact-match-wins, off/on word handling, and "unrecognised name selects nothing" are deliberate and each covered by a test.
- `/Ff` fix — Radio `0x8000` / Pushbutton `0x10000` now come from the shared `field_flags` constants, matching ISO 32000-1 Table 227.
- Tests are semantic (rasterize + ink measurement) with a real control and a save/reopen round-trip.

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the engine submodule is unchanged and still does not compile

Since last round nothing landed on the engine side: the submodule is still pinned at `3425288c`, so the compile error I flagged is still there. The Dart changes are clean. You also left a comment describing a changelog correction that is not in this push yet.

1. `vendor/pdf_oxide` (submodule `3425288c`) — `flush_form_fields_to_modified_objects` in `src/editor/document_editor.rs` passes `new_value` by value where `&FormFieldValue` is required. `button_on_state(new_value, &available)` (that fn takes `value: &FormFieldValue`) and `Object::from(new_value)` (only `From<&FormFieldValue> for Object` exists in `form_fields.rs`). Both need `&new_value`. I corrected my earlier note: `Some(val) => val.clone()` is fine — the errors are only those two call sites.
2. `CHANGELOG.pre.md` — the four new bullets are still under the released `## 4.1.0-dev.0` heading. There is no `## 4.2.0-dev.0`, and `CHANGELOG.md` has no `## 4.2.0`. Your comment describes exactly the right fix (new `4.2.0` headings on both lanes, `4.2.0` not `4.1.1` because of the Added bullet) and says it is verified — it just is not in `1a32bceb` yet, so it still needs to be pushed.

✅ **Checked**
- `setCheckboxFieldValue` is wired end to end: `BridgeEditorHandle` → `_SharedEditorHandle._mutate` → engine op, and the engine at `3425288c` dispatches `setCheckboxFieldValue`.
- The `setFormFieldValue` doc now states a button value is an appearance-state name, which is the correct mental model.
- The hand-authored fixture and semantic raster tests cover the reported cases and the off-word / non-`/Yes` / radio-kid paths coherently.

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the pinned engine commit still does not compile

Since the last round the changelog was corrected (new `4.2.0` / `4.2.0-dev.0` sections, both lanes, required-action → added → fixed order, MINOR bump for the added capability), but the blocking issue is unchanged.

1. `vendor/pdf_oxide:1` — the submodule is still pinned to `3425288c`. `flush_form_fields_to_modified_objects` in `src/editor/document_editor.rs` binds `new_value` as an owned `FormFieldValue` (`new_value: FormFieldValue = match &wrapper.modified_value { … }`) and then passes it by value at two call sites: `button_on_state(new_value, &available)` and `Object::from(new_value)`. `button_on_state` is declared `fn button_on_state(value: &crate::editor::form_fields::FormFieldValue, available: &[String])`, and `form_fields.rs` defines only `impl From<&FormFieldValue> for Object` — no `From<FormFieldValue>`. Both are E0308 at compile time, so `Rust tests` fails before anything runs. Fix is `&new_value` at both sites, then bump the submodule. The open thread from the previous round already covers this; it stays open.

✅ **Checked**
- Changelog now follows the repo convention: `4.2.0` in `CHANGELOG.md`, `4.2.0-dev.0` in `CHANGELOG.pre.md`, identical prose, and the two released `4.1.0*` entries left byte-identical.
- Dart API wiring is consistent end to end: `BridgeEditorHandle.setCheckboxFieldValue` → `_SharedEditorHandle._mutate('setCheckboxFieldValue', {fieldName, checked})` → `PdfEditor.setCheckboxFieldValue`, and the `setCheckboxFieldValue` op name is present in the engine dispatch.
- The `uncheckedButtonForm` fixture is self-consistent: 17 objects with matching cross-references, `/Ff 32768` = Radio, and each widget's `/AP /N` on-state filling 60×60 solid black while off leaves a hairline border.
- The band math is right: for a 792pt page, agree 700..760 → 0.04..0.12, optin 600..660 → 0.17..0.24, color 500..560 → 0.29..0.37, and the 0.05-on / 0.02-off dark-fraction cutoffs have clear margin against the ~9.8% a solid on-state produces.

slopfairy (dismissed):
> [!TIP]
> Approved — the value-as-state-name fix is correct end to end, and both earlier blocking flags are resolved

1. Engine `button_on_state` (`vendor/pdf_oxide`, `src/editor/document_editor.rs`) — an on-word (`'on'`, `'yes'`) on a radio group resolves per-kid in the flatten path (`modified_button_state`: each kid's sole state lights, so every kid turns on) but to `/Off` in the save path (`flush_form_fields_to_modified_objects`: the union of all kids' states has no sole state). The two paths disagree for `setFormFieldValue('color', 'on')`. Resolve against the group's union in the flatten path too, or document that on-words are checkbox-only.

✅ **Checked**
- My compile flag was wrong: `for (...) in &fields_to_flush` binds `new_value` as `&FormFieldValue` under match ergonomics (proven by `if *is_button` on the next line), so `button_on_state(new_value, …)` and `Object::from(new_value)` both type-check. No engine change needed there.
- Changelog: `4.2.0` and `4.2.0-dev.0` are new headings, the released `4.1.0`/`4.1.0-dev.0` entries are untouched, and `checks / Changelog Versions` is green.
- `/Ff` swap: `RADIO = 1 << 15` and `PUSH_BUTTON = 1 << 16` in `field_flags` are the correct ISO Table 227 values, and `annotations.rs` now reuses them instead of re-spelling swapped literals.
- `/V` and `/AS` are written as `Object::Name`, and `/AS` reaches each radio kid with siblings cleared to `/Off`.
- Dart: `BridgeEditorHandle` has exactly one implementor (`_SharedEditorHandle`), so the new abstract method breaks nothing.
- Fixtures and band math check out: `/Ff 32768` is radio, and the y/x fractions match the widget rects.

The stale compile thread is resolved — that flag was my misread, not a code defect.

PR #219, opened 2026-08-29
docs: point S8 at where release.sh actually lives

chaudharydeepanshu:
`docs/UPDATING.md` S8 tells you to preview the pub.dev changelog with:

```sh
bash tool/ci/release.sh --stamp-changelog vX.Y.Z
```

That path holds `release_hooks.sh` and `upgrade.sh` — nothing else. `release.sh` moved to the shared [whuppi/ci](https://github.com/whuppi/ci) repo, which CI reaches through the `release-tool` action (`whuppi/ci/actions/release-tool@v2.4.0` in `release.yml`). Following the recipe gives "No such file or directory".

Now names the real location, plus the two preconditions the old recipe left out: run it from the package root, and set `GITHUB_REPOSITORY` — the script exits immediately without it, which is not obvious from the error.

Verified by running the documented command and confirming it builds the filtered changelog.

**Also stale, not fixed here** — flagging rather than widening a docs PR:
- `docs/ARCHITECTURE.md:273` draws a `tool/ci/` tree containing `release.sh` and `reconcile_test_json.sh`; neither is there, and `release_hooks.sh` is missing from the drawing.
- `CHANGELOG.md:15` and `CHANGELOG.pre.md:15` cite `tool/ci/release.sh --check-versions` inside the changelog standard. Same wrong path, but these are published artifacts, so worth deciding on separately.

**Tests:** none — documentation. The assertion is that the documented command runs, which I checked.

**Breaking:** no.

### DIFF
--- a/docs/UPDATING.md
+++ b/docs/UPDATING.md
@@ -500,10 +500,18 @@ changelog" step of the upload-assets job — right before the publish
 approval gate. Open the workflow run, expand that step, and review
 the full changelog that will be published.
 
-To preview locally:
+To preview locally, run the same mode the publish job runs. `release.sh`
+lives in the shared [whuppi/ci](https://github.com/whuppi/ci) repo, not
+this one — CI reaches it through the `release-tool` action, so locally you
+point at a checkout of that repo (a sibling of this one). Run from this
+package's root, and set `GITHUB_REPOSITORY` — the script requires it while
+loading, whichever mode you ask for, and exits without it. Some modes want
+more: `--check-versions` also needs `BRANCH` (`dev` or `prod`, picking the
+lane it checks). `--help` lists every mode.
 
 ```sh
-bash tool/ci/release.sh --stamp-changelog vX.Y.Z
+GITHUB_REPOSITORY=whuppi/pdf_manipulator \
+  bash ../ci/tool/ci/release.sh --stamp-changelog vX.Y.Z
 cat CHANGELOG.md
 git checkout CHANGELOG.md   # restore
 ```


docs/UPDATING.md
```diff
@@ -500,10 +500,16 @@ changelog" step of the upload-assets job — right before the publish
 approval gate. Open the workflow run, expand that step, and review
 the full changelog that will be published.
 
-To preview locally:
+To preview locally, run the same mode the publish job runs. `release.sh`
+lives in the shared [whuppi/ci](https://github.com/whuppi/ci) repo, not
+this one — CI reaches it through the `release-tool` action, so locally you
+point at a checkout of that repo (a sibling of this one). Run from this
+package's root; the script reads `GITHUB_REPOSITORY` to resolve commit
```
slopfairy: "resolve commit links" misstates what the variable does. `--stamp-changelog` doesn't emit any commit links; `GITHUB_REPOSITORY` is consumed as repo identity for `gh --repo` calls and the install snippet, and for this mode it's only required to be set (the script's `REPO="${GITHUB_REPOSITORY:?…}"` line exits without it).

```suggestion
package's root; the script requires `GITHUB_REPOSITORY` and exits without
it (it feeds repo identity for gh calls and the install snippet). `--help`
lists every mode.
```

slopfairy (dismissed):
> [!TIP]
> Approved — the new S8 recipe points at the real script location and covers the two missing preconditions.

1. `docs/UPDATING.md:507` — change "reads `GITHUB_REPOSITORY` to resolve commit links" to "requires `GITHUB_REPOSITORY` and exits without it"; `--stamp-changelog` emits no commit links, and for this mode the value is only required to be set, not consumed.

✅ **Checked**
- `tool/ci/` contains only `release_hooks.sh` and `upgrade.sh`; no `release.sh` exists anywhere in this repo, so the old path was genuinely broken.
- `whuppi/ci/tool/ci/release.sh` exists and its `--stamp-changelog` mode overwrites the CWD's `CHANGELOG.md` after `require_tag`/`require_pubspec` — matches the recipe's `cat CHANGELOG.md` then `git checkout` restore.
- The script's `REPO="${GITHUB_REPOSITORY:?…}"` guard exits immediately without the var, confirming the "set it or it dies" precondition; `require_pubspec` enforces running from the package root.
- `release.yml` reaches the script via `whuppi/ci/actions/release-tool@v2.4.0` with `mode: --stamp-changelog`, so the "same mode the publish job runs" framing is accurate.
- The stale references flagged but not fixed here are real: `docs/ARCHITECTURE.md:273` draws `release.sh` and `reconcile_test_json.sh` (neither exists) and omits `release_hooks.sh`; `CHANGELOG.md:15` and `CHANGELOG.pre.md:15` cite `tool/ci/release.sh --check-versions`. Leaving those for a separate decision is the right call for a docs-scoped PR.

docs/UPDATING.md
chaudharydeepanshu: Right on both counts, and I have corrected it. `--stamp-changelog` emits plain `<sha> <subject>` lines with no links, so "resolve commit links" was wrong.

Went slightly wider than the suggestion because the requirement is not mode-specific: `REPO="${GITHUB_REPOSITORY:?…}"` sits at script top level (line 74), so it is a load-time precondition for every mode. The text now says that, and adds the neighbouring trap — `--check-versions` also needs `BRANCH`, which the changelog standard cites by name. Both verified by running them.

slopfairy: > @chaudharydeepanshu: "Right on both counts, and I have corrected it. `--stamp-changelog` emits plain `<sha> <subject>` lines with no links, so…"

The revision is correct. `GITHUB_REPOSITORY` is required at load time regardless of mode, and `--check-versions` additionally needs `BRANCH`; the wording in `docs/UPDATING.md` now states both accurately. Thread resolved.

slopfairy (approved):
> [!TIP]
> Approved — the corrected path and load-time preconditions match the script; verified against `whuppi/ci/tool/ci/release.sh` and `release.yml`.

✅ **Checked**
- `release.sh` is genuinely gone from this repo — `tool/ci/` holds only `release_hooks.sh` and `upgrade.sh`. The script lives at `whuppi/ci/tool/ci/release.sh` (fetched and read it).
- The `GITHUB_REPOSITORY` wording is accurate: `REPO="${GITHUB_REPOSITORY:?...}"` sits at script top level, before any mode dispatch, so it is a load-time requirement for every mode — not specific to `--stamp-changelog`.
- `--check-versions` also requires `BRANCH`: `cmd_check_versions` uses `${BRANCH:?…}` and maps `dev`→`CHANGELOG.pre.md`, `prod`→`CHANGELOG.md`, matching "picking the lane it checks".
- The command shape is correct: `--stamp-changelog` needs a `v`-prefixed tag and a `pubspec.yaml`, and that mode never calls `gh` (curl to pub.dev + `git log` only), so the docs correctly omit any `GH_TOKEN` requirement for a local dry run.
- `release.yml` reaches the script through `whuppi/ci/actions/release-tool@v2.4.0` with `mode: --stamp-changelog` in the publish job, matching "the same mode the publish job runs".
- The three stale citations you flagged out of scope are all real: `docs/ARCHITECTURE.md:273` lists `release.sh` and `reconcile_test_json.sh` (neither present, and `release_hooks.sh` missing from the drawing); `CHANGELOG.md:15` and `CHANGELOG.pre.md:15` cite `tool/ci/release.sh --check-versions`. Correctly left for a separate decision — those two are published artifacts.

PR #213, opened 2026-07-30
ci: bump all action refs (workflows + composites)

chaudharydeepanshu:
One owner for every `uses:` action ref in `.github` — workflow and
composite alike, no split with Dependabot:

- **whuppi/ci refs** across `.github` (workflows AND composite
  `action.yml`) swept uniform to the latest whuppi/ci release, so the
  pin never splits.
- **Every third-party action** (workflows AND composites) pinned to the
  latest SHA via pinact — including the refs inside composite
  `action.yml` that Dependabot can't read (dependabot-core#6704).

Dependabot owns pub deps only.

Add `ready-to-test` for the full cross-target run, then merge once green.

_Auto-generated by upgrade-check.yml_

PR #220, opened 2026-08-29
ci: bump whuppi/ci refs to v2.4.1

chaudharydeepanshu:
Picks up [whuppi/ci#43](https://github.com/whuppi/ci/pull/43), released as [v2.4.1](https://github.com/whuppi/ci/releases/tag/v2.4.1).

## Why this one matters here

`Upgrade Check` has been red on `dev` since **6 August** — a 22-run streak. The `action-refs` job's staging loop ended in `[ -n "$f" ] && git add -- "$f"`. An empty `PINNED_FILES` still contributes a blank line to the `printf`, so the last iteration's test is false, `&&` short-circuits to exit 1, that becomes the loop's status, and the loop is the right side of a pipeline — so `set -e -o pipefail` killed the step before it could commit, push, or update the PR.

The sweep had already rewritten the refs correctly on every run; only staging fell over, so the work was thrown away silently. That is why #213 has been frozen at `f52b9a8` — the commit from the last run that happened to have pin changes (6 Aug, `PINNED_FILES: .github/workflows/ci.yml`).

## What's in v2.4.1

- The `action-refs` fix above.
- Pinned tools: fvm 4.1.2 → 4.3.0, zizmor 1.28.0 → 1.29.0, pinact 4.1.0 → 4.1.1, Chrome 151 → 152.
- Pinned actions: checkout v7.0.1, labeler v7.0.0, setup-java v5.7.0, stale v11.0.0.

## This PR

All **30** refs across the 9 workflows and the vendored `make-target` composite, moved together — the pin-availability gate rejects a split, which is the whole reason the sweep exists. Verified uniform afterwards: 30 at `v2.4.1`, 0 left at `v2.4.0`, and no stray refs outside `.github/`.

No changelog entry: CI ref bumps aren't changelog events here (`f52b9a8`, the last such bump, touched neither file).

Once this lands, the next daily radar should go green and #213 should start updating again on its own.

**Tests:** none — CI configuration. The assertion is the `Pin availability` required check, which verifies the pinned whuppi/ci release exists.

**Breaking:** no.

### DIFF
--- a/.github/actions/make-target/action.yml
+++ b/.github/actions/make-target/action.yml
@@ -75,7 +75,7 @@ runs:
     # the native/wasm ones (rust, wasm-cache, wasm-build) are pdf-specific and
     # stay local. The shared capabilities were generalized FROM these, so the
     # behavior is identical — only their home moved.
-    - uses: whuppi/ci/actions/capabilities/fvm@v2.4.0
+    - uses: whuppi/ci/actions/capabilities/fvm@v2.4.1
 
     # ── On demand ──
     # Default on: most jobs build the native/wasm engine. Dart-only static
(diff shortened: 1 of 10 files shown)


slopfairy (approved):
> [!TIP]
> Approved — clean lockstep bump of every whuppi/ci ref to a published v2.4.1.

✅ **Checked**
- `whuppi/ci` release `v2.4.1` exists and is published (tag `v2.4.1`, `aa6e3a6`, 2026-08-29) — the pin-availability gate will pass, and its release notes match the `action-refs` staging-loop fix described in the body.
- All 30 `uses: whuppi/ci/...` refs move `v2.4.0` → `v2.4.1` together across the 9 workflows plus `.github/actions/make-target/action.yml` (12 there, 10 in release.yml, 1 in each of the other 8). Every removed line is `@v2.4.0` and every added line `@v2.4.1`; additions and deletions both total 30 — no split, no stragglers.
- `ci.yml` is correctly untouched: it references only the local `./.github/actions/make-target` and SHA-pinned external actions (`actions/checkout`, `dorny/paths-filter`), with no whuppi/ci refs to bump.
- The local capability actions (`rust`, `wasm-build`, `wasm-cache`) contain no whuppi/ci refs, so nothing outside the 10 changed files was missed.
- No rolling refs: every target is a pinned semver tag (`@v2.4.1`), none moved to `@main`/`@stable`.
- Diff touches only `.github/` files, so the "no changelog entry" claim holds — neither `CHANGELOG.md` nor `pubspec.yaml` is in the changeset.

PR #248, opened 2026-09-12
fix: DOCX tables convert at their declared widths (#243); vendors synced; web worker budget frees a slot on confirmed death

### LINKED ISSUE #243: office: convertToPdf() ignores w:tblGrid/w:tblW — DOCX table rendered ~3.4x its declared width, columns pushed off page
## Summary

`convertToPdf()` (DOCX → PDF, `office` capability) appears to ignore a table's declared column widths entirely when laying out the PDF — it renders the table roughly **3.4x wider than both the page and the document's own `w:tblGrid`/`w:tblW`**, running most columns off the page. This isn't a "table too wide for the page" authoring issue on the source document's side: I patched the source `.docx`'s width metadata directly and the rendered output didn't change at all, which points to the layout engine not reading those fields for this code path.

## Environment

- `pdf_manipulator: 4.1.0` (latest as of testing)
- Reproduced on Windows (Dart standalone script, native-assets build hook), and separately on Android via the Flutter app this is embedded in
- `dart --version`: whatever ships with Flutter 3.47.1 / Dart 3.13.1

## Minimal repro (numeric spec — no attachment needed to reproduce)

Source `.docx` section/table properties:

```xml
<w:pgSz w:w="15840" w:h="12240" w:orient="landscape"/>
<w:pgMar w:top="720" w:right="630" w:bottom="810" w:left="810" w:header="720" w:footer="720" w:gutter="0"/>
...
<w:tblLayout w:type="fixed"/>
<w:tblW w:w="0" w:type="auto"/>
<w:tblGrid>
  <w:gridCol w:w="666"/>
  <w:gridCol w:w="1330"/>
  <w:gridCol w:w="1237"/>
  <w:gridCol w:w="1401"/>
  <w:gridCol w:w="1258"/>
  <w:gridCol w:w="1236"/>
  <w:gridCol w:w="2160"/>
  <w:gridCol w:w="1440"/>
  <w:gridCol w:w="1260"/>
  <w:gridCol w:w="1440"/>
  <w:gridCol w:w="1368"/>
</w:tblGrid>
```

That's an 11-column table, `gridCol` widths summing to 14,796 twips (10.275in), on a landscape page whose usable width (page − margins) is 14,400 twips (10in) — i.e. the table as authored is only ~2.6% over the printable width, essentially a non-issue in any normal renderer (opens fine in Word/LibreOffice at ~10.3in). Each cell holds a paragraph of ordinary body text (not one long unbroken word/run — normal prose, word-wraps fine in Word).

Code:

```dart
final pdf = Pdf();
final sink = await FileSink.create(outputFile);
await pdf.convertToPdf(FileSource(inputDocxFile), sink, format: PdfDocumentFormat.docx);
await sink.close();
```

## Actual result

Output PDF page (`MediaBox`) is `[0, 0, 792, 612]` (792pt = 11in, matching the landscape page). Extracting text run positions (via `pypdf`'s `extract_text(visitor_text=...)`, reading each run's text-matrix x/y) shows table content placed at x-coordinates up to **2543.6pt** — over 3.2x the page width, and about 3.4x the table's own declared 10.275in width. Most columns are placed entirely outside the visible page area.

## What I ruled out

1. **Table too wide as authored** — no: `gridCol` sum is ~10.3in against a ~10in usable page width, a trivial overage that any reasonable renderer would either honor as-is (slightly over margin) or clip by ~0.3in, not balloon to 35+in.
2. **Explicit width metadata not being read** — tested directly: rewrote `w:tblW` from `w:type="auto" w:w="0"` to an explicit `w:type="dxa" w:w="14400"`, and proportionally rescaled every `w:gridCol` and `w:tcW` to sum exactly to the 14400-twip usable width. Reconverted with the same `convertToPdf()` call. **Result: byte-for-byte identical text-run x-positions** (max x still 2543.592). The explicit widths had zero effect.
3. **Auto-sizing from unwrapped content length** — tested directly: inserted `<w:br/>` manual line breaks into every table-cell text run longer than 30 characters (so no cell has an unbroken run wider than ~30 chars), reconverted. **Result: max x-position was still ~2543–2559pt**, no meaningful change. So it's not sizing columns off the longest unbroken text run either.

Both of these were real edits to `word/document.xml` inside the `.docx`, each individually reconverted and re-measured — not speculation. Neither had any effect on the rendered column layout, which suggests the DOCX table-layout path for `office`/`convertToPdf` is computing column widths from something other than the document's own `tblGrid`/`tblW`/cell content — possibly a fallback/default sizing path that isn't being overridden by any of the standard OOXML mechanisms for expressing column width.

## Expected

Table columns should be laid out at (or proportionally fit to) the widths the source document specifies via `w:tblGrid`/`w:tblW`, clipped or scaled to the destination page size if they exceed it — not rendered at ~3.4x that width regardless of what the document declares.

## Notes

- I can't attach the original document (contains a real school name and a real person's name I'm not able to publish), but the numeric spec above (page size, margins, exact `gridCol` values, `tblLayout`/`tblW` settings) is everything needed to build an equivalent minimal repro from scratch — happy to also generate and share a fully synthetic `.docx` with placeholder text reproducing the same structure if that'd help, just let me know.
- Single-table, single-section document; no merged cells (no `gridSpan`/`vMerge` present), so this isn't a merged-cell edge case.

chaudharydeepanshu:
Fixes #243.

## The bug
`convertToPdf()` laid a DOCX table out at ~3.4x its declared width. The converter (`vendor/pdf_oxide`, `render_table_chunk`) sized every column by sampling cell content with a 300 pt cap and never read the widths the document declares in `w:tblGrid` — which is why no edit to `w:tblW`, `w:gridCol` or `w:tcW` changed anything for the reporter. office_oxide had parsed the widths into the IR all along (`column_widths_twips`); the converter ignored them.

**Fix:** declared widths drive fixed-mode columns, scaled down proportionally when their sum exceeds the usable page width. Content sampling remains only for tables that declare nothing, with the per-column cap bounded so a chunk can never outgrow the page.

**Proof:** `test/fixtures/handwritten_docx.dart` builds a DOCX from the issue's exact numeric spec (eleven `gridCol`s summing to 14,796 twips on a 15,840-twip landscape page). Before the fix the converted row spanned to x = 3073.8 pt on the 792 pt page; after, every column marker sits inside it. The test runs on native and all three web lanes.

## Vendors synced (UPDATING.md S1)
- pdf_oxide v0.3.73 → **v0.3.78**, 31 patches rebased. Upstream landed the same R6 encryption-key fix we carried, so that patch is dropped; our `/Perms` fix (R6 requires it, upstream still writes `None`) rides on top. Upstream's per-page redaction function now delegates to our explicit-regions core. Two API drifts adapted (`render_form_xobject` clip/state args; x509-cert accessors).
- office_oxide v0.1.3 → **v0.1.11**, 2 patches rebased.
- `make test-rust` green for both; fork branches renamed, mirrors and tags synced; `build.json` baseTag and the UPDATING.md fork table updated.

## Web worker budget frees a slot on confirmed death
The full check made the known "5000x dispose" Chrome OOM deterministic. Root cause, not a flake: the page-wide worker budget freed a slot the moment `worker.terminate()` was called, but the browser releases a worker's WASM memory later, so rapid create+dispose churn booted new workers into memory that dead ones still held (`Cannot allocate Wasm memory for new instance`). Every worker now holds a Web Lock from before its WASM instance exists until it dies, in all three I/O modes (OPFS reclaim already used it), and a slot returns only when that lock is acquirable — the same rule the native lane table follows. `dispose()` stays instant; only a boot past the cap waits its turn. No timer, no retry. ARCHITECTURE.md's "Kill semantics (web)" and "Web global budget" rows updated.

## whuppi/ci 2.7.4
Refs bumped; the four stamped gate scripts re-stamped. `lint_shell.sh` now rejects inline lint suppressions fleet-wide, so the two `zizmor: ignore[dangerous-triggers]` in `retry.yml`/`triage.yml` move to `.github/zizmor.yml` as one reviewed policy with its reason.

## Changelog
`CHANGELOG.pre.md` — `4.2.1-dev.0`.

## Verification
`make check` green on the whole tree (analyze, native, OPFS, JSPI, atomics, example web ×3); the atomics lane run three more times in a row after the budget fix, 251/251 each.


slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — the zizmor ignore paths won't match the workflows, so the lint gate this PR is trying to keep green regresses

1. `.github/zizmor.yml:24` — the `dangerous-triggers.ignore` entries are bare filenames. zizmor glob-matches `ignore` entries against the repository-relative path, so `retry.yml` / `triage.yml` never hit `.github/workflows/retry.yml` / `.github/workflows/triage.yml`. The inline suppressions this PR removes were the working mechanism; with them gone and the config not matching, the gate flags both workflows again. Full paths fix it (inline suggestion posted).

1. `.github/workflows/triage.yml:8` — the header still points at "the ignore on the trigger below", which no longer exists. retry.yml's header was updated to reference `.github/zizmor.yml`; this one wasn't.

✅ **Checked**
- `lib/src/runtime/web/host.dart` budget accounting: every failure path returns its slot exactly once. Spawn throw decrements directly; boot failure routes through `_terminateAndReleaseSlot` → `_releaseSlotWhenDead` → `finally` decrement; `_prewarm`'s catchError correctly relies on `_bootWorker`'s own failure path. No double-free, no leak.
- The deferred-release ordering is sound: `_releaseSlotWhenDead` decrements and wakes a waiter only after the lock is acquired, so a woken `takeWorker` can never boot past the cap while dead WASM memory is still held.
- `web_assets/lane_worker.js` acquires the liveness lock before the WASM instance exists and holds it until death; the host uses lock acquisition as the death certificate, the same signal OPFS reclaim already keyed on. Non-secure-context fallback (no `locks`) resolves immediately, matching the documented backstop.
- The DOCX fixture is a valid hand-authored OOXML package, and the test asserts each column marker's line right edge stays inside the 792 pt page — it catches the ~3074 pt regression and runs on native plus all three web lanes.
- `archive` is dev-only (`direct dev`, 4.0.9 in the lock), used only as an independent fixture producer, never the package's own writer.
- Shell rewrites (`read -ra` + array-or-nothing expansion) are bash 3.2-safe; `verify_web_gate.sh` guards `FLUTTER` with `:?` before splitting, so the unguarded `"${flutter_cmd[@]}"` is safe.
- CI ref bumps (v2.7.3 → v2.7.4), `build.json`, `UPDATING.md` fork table, and `CHANGELOG.pre.md` are internally consistent.

slopfairy (dismissed):
> [!TIP]
> Approved — the worker-budget fix is sound and the fixture reproduces issue #243 exactly.

1. `test/ops/core/pdf_standalone_battery.dart:172` — the assertion proves only that the columns stay on the page (right edge ≤ 793 pt), which the content-sampling fallback would also satisfy. Add a check on the markers' left edges following the declared `docxWideTableGridCols` proportions (monotonically increasing, first near the 40.5 pt left margin, last near the 760.5 pt usable right edge) so a wrong-but-on-page layout fails.

✅ **Checked**
- `lib/src/runtime/web/host.dart` — traced every `_liveBudget` increment/decrement path: `takeWorker`, `_prewarm`, `_bootWorker`'s success and both failure branches, and `_releaseSlotWhenDead`'s `finally`. No double-decrement and no slot leak; failed boots return their own slot.
- The boot-failure branch passes `opfsDir: null`, which is correct: `laneDir()` in `web_assets/lane_worker.js` is created lazily only under work, and work requires `ready`.
- `web_assets/lane_worker.js` — the liveness lock is acquired before `ensureInit()` (WASM instantiate) in all three I/O modes and held until agent death; the host acquires it as the death certificate.
- Workflow security — `triage.yml` keeps `pull_request_target` under `contents: read` with no PR-head checkout; `retry.yml`'s only fork-influenced value is the numeric run id. The zizmor suppression is centralized in `.github/zizmor.yml` with a reason.
- `tool/lint_shell.sh` — the new inline-suppression scan matches no tracked source file (only Markdown prose and the script itself, both excluded), so it won't false-positive.
- `pubspec` — `archive` resolves to 4.0.9 under `^4.0.9`, dev-only scope correct; the v2.7.3 → v2.7.4 CI bumps are mechanical and consistent.

slopfairy (dismissed):
> [!TIP]
> Approved — the worker-budget liveness-lock fix is correct, and the DOCX fixture faithfully reproduces issue #243's numbers.

1. `test/ops/core/pdf_standalone_battery.dart:204` — `lessThanOrEqualTo(pageWidthPt + 1)` only fails on overflow, so a table that now renders too narrow would still pass. Bound the rightmost marker's right edge from below too (the columns sum to ~740 pt, ~720 pt after proportional scaling), so the test catches over-shrinking, not just the 3.4x blowup it guards today.
2. `test/fixtures/handwritten_docx.dart:17` — change "2.6% over" to "2.75% over": 14,796 / 14,400 = 1.0275.

✅ **Checked**
- `lib/src/runtime/web/host.dart` — slot accounting is balanced across every boot-failure path (`_spawnWorker` throw decrements inline; the handshake-failure path decrements in `_releaseSlotWhenDead`'s `finally`), `worker.js.terminate()` now happens only inside `_terminateAndReleaseSlot`, and FIFO wake ordering is preserved.
- `web_assets/lane_worker.js` `acquireLivenessLock` holds the lock from before WASM init until agent death, matching the host now passing `livenessLock` in all three I/O modes; the worker file needs no change, which is consistent with it being absent from the diff.
- `tool/analyze_core.sh`, `platforms_gate.sh`, `verify_web_gate.sh` — `read -ra` splits without pathname expansion and the `${arr[@]+"${arr[@]}"}` guard is `set -u`-safe on bash 3.2.
- `tool/lint_shell.sh` check 6 — no false positives in the tree: the only non-excluded tracked matches are Markdown (`AGENTS.md`, `CONTRIBUTING.md`), which the scan skips; the former `zizmor: ignore[dangerous-triggers]` lines are gone from `retry.yml`/`triage.yml`.
- `.github/workflows/triage.yml` / `retry.yml` — `pull_request_target` and `workflow_run` only delegate to the hardened reusable workflow: least-privilege per-job permissions, no fork checkout, event values confined to `concurrency` groups (never `run:`), refs pinned to `@v2.7.4`. The zizmor `dangerous-triggers` ignore in `.github/zizmor.yml` is sound under that posture.
- Vendor sync is internally consistent: `build.json` baseTag v0.3.78, `UPDATING.md` fork branches/tags, and both submodule bumps agree.

slopfairy (dismissed):
> [!TIP]
> Approved — the worker-budget slot accounting is correct, and the DOCX-width regression is covered by a fixture built from the issue's exact numbers

Re-reviewed the full changeset against `dev`. No prior inline threads are open, so nothing to resolve. The budget rework is the piece that needed the closest read; it holds up.

1. `lib/src/runtime/web/host.dart:381` — the "a failed boot returns its own slot" guarantee now lives inside `_bootWorker`, but the span between `_spawnWorker()` and the `try` at line 416 (completer setup plus both `EventStreamProviders.*.listen(...)` calls) sits outside any slot-releasing handler. The old `takeWorker` wrapped the whole `_bootWorker()` call in try/catch, so it covered this window; now it does not. Moving the listener setup inside the existing `try`, or wrapping the whole body in a `try/finally` that decrements `_liveBudget` on the way out, restores the invariant that any boot failure returns its slot. Narrow, but the invariant is what this PR is about.

✅ **Checked**
- `lib/src/runtime/web/host.dart` — traced slot accounting across `takeWorker` / `_prewarm` / `returnPristine` / `retire`. Each worker reserves one `_liveBudget` slot and releases it exactly once; the spawn-fail, handshake-fail, pool-overflow, and retire paths all reconcile.
- The liveness lock is the same death signal on both sides: the worker holds it from before `ready` (`web_assets/lane_worker.js` `acquireLivenessLock`), and `_releaseSlotWhenDead` frees the slot only when it can acquire the name — matching the documented native-lane semantics. Non-secure-context fallback (no `navigator.locks`) frees the slot immediately on both sides, as documented.
- `tool/lint_shell.sh` check 6 — the suppression regex excludes Markdown and the script itself, and a repo-wide grep confirms no remaining `// ignore:` / `shellcheck disable=` / `noqa` markers in tracked source. The three `read -ra` word-splitting rewrites are bash 3.2-safe.
- `.github/zizmor.yml` `dangerous-triggers` policy and the `retry.yml` / `triage.yml` edits — both workflows keep least-privilege job tokens, no checkout of fork code, and no `github.event.*` interpolation into run blocks. All `whuppi/ci` refs are exact `@v2.7.4` tags, not rolling.
- `archive` dev dependency — `^4.0.9` in pubspec matches `4.0.9` under `direct dev` in the lock; dev-only, ships nowhere.
- `test/fixtures/handwritten_docx.dart` geometry and the new assertion — the twip sum is 14,796 as stated, and the single-table-wide scale factor with 3% tolerance is a sound way to reject a content-sampled (equal-gap) layout rather than merely an on-page layout.

slopfairy (dismissed):
> [!TIP]
> Approved — the worker-budget liveness-lock fix is sound, and the DOCX-width test now asserts the declared-width geometry rather than page containment alone

One commit since the last round: the `whuppi/ci` refs move v2.7.4 → v2.7.5 (sdkmanager launcher matched by `bin/` path). I grep-verified every `whuppi/ci` ref in the tree sits on `@v2.7.5`, so the pin-availability gate holds. All prior-round suggestions landed in earlier commits: the geometry assertion (proportions plus row-span lower bound), the 2.75% fixture comment, and the triage.yml header now pointing at `.github/zizmor.yml`.

1. `lib/src/runtime/web/host.dart:381` — carry-over, unchanged since last round: the completer setup and both `EventStreamProviders.*.listen(...)` calls sit between the `_spawnWorker()` try/catch and the main `try`, outside any slot-releasing path. A throw there would leak the reserved `_liveBudget` slot; folding them into the existing `try` restores the "failed boot returns its own slot" invariant.

✅ **Checked**
- Slot accounting is balanced: `takeWorker`/`_prewarm` reserve once; spawn-fail, handshake-fail, `retire`, and pool-overflow `returnPristine` each release once via `_releaseSlotWhenDead`'s `finally`. No double-free, no leak.
- Deferred release is correct: a slot and a waiter are released only after the worker's lock is acquired, so a woken taker cannot boot past the cap while dead WASM memory is still held; non-secure contexts degrade to immediate release on both host and worker.
- `web_assets/lane_worker.js` holds the liveness lock from before `ensureInit()` until agent death in all three I/O modes; the host keys budget and OPFS reclaim on the same lock grant.
- The DOCX test rejects both failure modes: per-column gaps must scale to one table-wide factor (3% tolerance, kills the content-sampled equal-gap layout), and the row must span at least 75% of the page (kills over-shrink). The fixture reproduces issue 243's exact 14,796-twip spec.
- The three `read -ra` + `${arr[@]+"${arr[@]}"}` rewrites are `set -u`-safe on bash 3.2 and `set -e`-safe (the herestring always delivers a full line).
- Workflow security holds: `triage.yml`/`retry.yml` keep least-privilege per-job tokens, no fork checkout, event values confined to `concurrency` groups; the `dangerous-triggers` exception is centralized in `.github/zizmor.yml` with its reason, and `tool/lint_shell.sh` check 6 matches no tracked source file.

slopfairy (dismissed):
> [!TIP]
> Approved — the last carry-over is fixed, and the zizmor dangerous-triggers policy is confirmed working by the CI gate

Since round 5, one commit landed (`41129d63`): the completer setup and both worker listeners moved inside the slot-owning `try`, and the `finally` only cancels subscriptions once they exist. That closes the leak window I had flagged twice. Nothing else in the tree changed.

✅ **Checked**
- `lib/src/runtime/web/host.dart` — re-traced every `_liveBudget` path on the new shape: spawn-throw decrements inline; any throw after the spawn routes through `_terminateAndReleaseSlot` → `_releaseSlotWhenDead`, whose `finally` decrements exactly once; success holds the slot until `retire` or pool-overflow. No double-free, no leak. The `LaneWorker(worker, null, lockName)` in the boot-failure catch is correct — a never-readied worker wrote no OPFS directory (writing requires work, work requires ready) but may still hold its liveness lock, so the host still waits for the lock before returning the slot.
- `.github/zizmor.yml` — the bare `retry.yml` / `triage.yml` ignore entries work. The Workflow lint job on the head commit ran zizmor 1.29.0 and reported "No findings to report (2 ignored)" — exactly the two privileged-trigger workflows. That resolves my first-round concern: zizmor matches the entries against the workflow basename, so the centralized policy does suppress both findings, and the "2 ignored" count proves the rule is active rather than disabled.
- Everything else carries over unchanged from the prior round and still holds: the DOCX test asserts both the declared-width scale factor and the row-span lower bound; vendor sync is internally consistent (`build.json` v0.3.78, `UPDATING.md` fork table, both submodule bumps); the `read -ra` + array-or-nothing rewrites are bash 3.2-safe; changelog entries match the pre-release version.

slopfairy (approved):
> [!TIP]
> Approved — the DOCX declared-width fix is real and pinned by a geometry-checking regression test, and the worker-budget change returns a slot only on confirmed death, end to end.

✅ **Checked**
- `lib/src/runtime/web/host.dart` — traced every `_liveBudget++` (take + prewarm) to exactly one release: spawn-throw decrements inline, boot-throw and retire/returnPristine terminate via `_terminateAndReleaseSlot`, and `_releaseSlotWhenDead` decrements in `finally`. No double-free, no leak. The 41129d63 fix does what it claims — the subscription setup and both completers now sit inside the slot-owning try, so a throw after `_spawnWorker` still terminates the worker and frees the slot.
- `web_assets/lane_worker.js` — `acquireLivenessLock` holds a never-resolving Web Lock from before the WASM instance exists until the agent dies, in all three I/O modes. `terminate()` → browser reaps → lock frees → the host's queued `request` grants. The death signal is real, not a timer.
- `test/fixtures/handwritten_docx.dart` + the new battery test — independent producer (byte-built OOXML via `archive`, dev-only), asserts on-page, per-column geometry (the 666-twip first column would expose an equal-gap sampled layout), and ≥75% page span. Correct and tight.
- `.github/` — every whuppi/ci ref sits on the exact tag v2.7.6; no fork-influenced value is interpolated into a `run:` block; the two `zizmor: ignore` lines moved to one reviewed `.github/zizmor.yml` policy. Fail-safe: a non-matching ignore re-flags the trigger, it cannot silently skip.
- `tool/` — the `read -ra … <<<` word-split refactors are bash-3.2-safe; `lint_shell.sh` section 6 excludes markdown and itself, and the only remaining `// ignore:` / `# noqa` text in the tree lives in AGENTS.md/CONTRIBUTING.md (excluded).
- Vendor sync — `build.json` baseTag v0.3.78 and the UPDATING.md fork table (`pdf_manipulator/0.3.78-patches`, `office_kit/0.1.11-patches`) agree; the engine change is validated by the regression test rather than by reading 31 rebased Rust patches.
- CI — full-test gate plus compile/verify/int/pkg across the matrix are green; only the slow Android emulator legs were still running at check time.

PR #249, opened 2026-09-12
feat!: 5.0.0 — image reducer, image edits, and every PLANNED capability row

chaudharydeepanshu:
Image editing and image reduction for the editor, one writer bug that made the old optimizer a no-op, and the review finding on `pageImages`. **Major bump (5.0.0): breaking changes listed first.** Also closes every PLANNED capability row in the roadmap (26 rows) — see §4.

## Breaking
- `flattenAllAnnotations()` → `flattenAnnotations({int? page})`; `flattenForms()` gains the same optional `page`.
- `applyRedactions()` returns a `PdfRedactionReport`.
- `mergeFrom(other)` gains `pages:`; `embedFile(name, data)` gains `description:`, `mimeType:`, `relationship:`.
- `PdfEditor.optimizeImages(quality:, minSize:)` is removed. `reduceImages(PdfImagePolicy)` is the one image reducer; the old behaviour is `reduceImages(PdfImagePolicy(jpegQuality: q, minPixels: m, convertCmykToRgb: true))` and the old count is `report.changed`.
- `compress(source, output, imageQuality:)` is now `compress(source, output, images:)`, a `PdfImagePolicy` defaulting to `PdfImagePolicy.ebook` (150 ppi, JPEG q75). A plain `compress` now downsamples images drawn above 225 ppi.
- Editor and doc reads without an argument are `PdfTask` getters (a property reads as a noun, never with a `get` prefix): `getTitle()` → `title`, `getAuthor()` → `author`, `getSubject()` → `subject`, `getKeywords()` → `keywords`, `getProducer()` → `producer`, `getCreationDate()` → `creationDate`, `PdfDoc.getSignatures()` → `signatures`. The one read with an argument is a noun method: `getPageMediaBox(page)` → `pageMediaBox(page)`, matching `pageImages(page)`.

## 1. `pageImages`, `repositionImage`, `setImageBounds`
`PdfEditor.pageImages(page)` lists every image XObject drawn on a page with its resource name, placement bounds and full transform. That name is what `resizeImage` (existing), `repositionImage` and `setImageBounds` (new) take, so the three edits share one vocabulary and each is proven by listing again after save. Engine fix on the way: the edits wrote absolute values into the image's own `cm` while the listing reports the composed matrix, so an image under an enclosing transform moved to `x + offset`; they now compose.

Review finding (slopfairy, fixed): upstream's walker records every `Do`, forms included. `pageImages` now filters through the page's `/Resources /XObject` (inherited resources too) to `/Subtype /Image`. A handwritten fixture draws an image and a Form XObject on one page: red before, green after.

## 2. `reduceImages(PdfImagePolicy)` — a policy-driven image reducer
The old optimizer only touched 8-bit Device* Flate images, never JPEG, never resampled, and ignored how large an image is drawn; on a scanned or photo PDF it did nothing. It is replaced by `vendor/pdf_oxide/src/host/images/`, five stages:

```
Inventory ──► Classify ──► Decide ──► Execute ──► Report
```

- **Inventory** groups every painted image XObject by object across the visible pages (upstream's `page_image_handles`, forms composed in), one placement per `Do`.
- **Classify** reads the dictionary (a staged replacement from earlier in the session wins): encoding (raw / jpeg / jpx / ccitt / jbig2), colour model (bilevel / gray / rgb / cmyk / other; Indexed resolved to its base, ICCBased by N), bit depth, soft mask, stencil or colour-key mask.
- **Decide** is a pure table over kind × placements × policy (unit-tested): effective ppi from the CTM, downsample only above target × 1.5 and then to the target, a JPEG never re-encoded in place, CMYK → RGB only when asked, lossless never picks a lossy codec.
- **Execute** decodes through upstream's extractor (Indexed expanded, `/Decode` folded, 16-bit collapsed, Adobe CMYK JPEG handled), resamples (Triangle; bilevel area-average + threshold), encodes (JPEG; Flate + PNG predictors; CCITT G4 via `fax`), rewrites the dictionary, co-processes the `/SMask`, and stages the replacement only when parent + mask are smaller.
- **Report** returns one typed row per image: what it was, what happened, why it was kept.

Dart: `PdfImagePolicy` with `screen` / `ebook` / `print` / `lossless` presets and the knobs `colorPpi/grayPpi/monoPpi`, `downsampleThreshold`, `jpegQuality`, `chromaSubsampling`, `minSavings`, `recompressJpeg`, `allowLossy`, `convertCmykToRgb`, `minPixels`; `PdfImageReport` / `PdfImageOutcome`; `compress(images:)`. Measured: `images_hires` (128² photo drawn at 288 ppi) 49,409 → 3,479 bytes under `screen`.

## 3. Writer bug: staged image replacements were dropped on save
`write_full_to_writer`'s resource walk copied every XObject and ExtGState referenced from a page straight from the source, and its final sweep skips ids already written. A staged image never reached the file (its soft mask did, since the mask is only reachable through the image). The old `optimizeImages` reported a count and saved the original bytes since it shipped; its test only checked the count. The walk now goes through `staged_or_source_object()` (S2 patch).

## 4. Roadmap sweep — every PLANNED capability row → DONE
One layering rule, written into the roadmap: *PdfDoc answers questions about a document as it is; PdfEditor answers questions about the session's staged state; exports live on PdfDoc or PdfStandalone.*

- **Page geometry (editor):** `pageCropBox`, `setPageMediaBox`, `setPageCropBox`, `setPageRotation` (absolute).
- **Forms:** `PdfDoc.formFields` / `formField(name)` (typed `PdfFormField` with a sealed value), twelve property setters on the editor (`removeFormField`, read-only, required, tooltip, bounds, max length, alignment, background/border colour, border width, appearance, flags — all typed, no raw PDF syntax), `PdfDoc.exportFormData` (FDF / XFDF).
- **Scoped edits:** `flattenForms(page:)`, `flattenAnnotations(page:)`, `clearEraseRegions`, `sanitize(PdfSanitizeOptions)`.
- **XFA:** `PdfDoc.xfa` (`PdfXfaInfo?`), one-shot `convertXfaToAcroForm`.
- **Selective merge:** `mergeFrom(other, pages:)` (engine: `merge_pages_from_reader`, S2 patch).
- **Attachments:** `PdfDoc.attachments` (name-tree walk, no stream decode), `extractAttachment`, `embedFile` options — which also closes the "attachment-listing" gap row.

Engine bugs the new batteries forced into the open, all fixed at the cause (S2-marked where upstream): form property setters only persisted through the flatten path and `removeFormField` never removed (write path now writes every dirty key and drops deleted fields); `mergeFrom` dropped the merged document's AcroForm; `pageMediaBox` returned the far corner as size; `addStamp`, the builder's `image`, `eraseRegions` and `addRedaction` passed sizes where corners were expected (`eraseRegions` has erased the wrong rectangle since it shipped); an embedded file's MIME type was escaped twice.

Remaining PLANNED rows are infrastructure (typed wire error codes, item streaming, memory verification, builder hardening, an annotation-listing read API, and the reducer's appearance-stream walk) — each its own PR.

## Proof
- Rust: 20 unit tests in `host/images/` — the decision table; Flate-predictor, JPEG and CCITT G4 round trips through the crate's own decoders; resampling properties — each shown able to fail.
- Fixtures from three independent producers: dart-pdf (`images_hires`, `images_jpeg`, `images_smask`), ImageMagick (gray / CMYK JPEG / Indexed / CCITT G4, committed with the exact commands), handwritten ASCIIHex micro fixtures (1-bit raw, 16-bit, stencil mask, Separation, colour-key mask, JBIG2 stub, one XObject shared by two pages, an image inside a form, inherited resources).
- Editor battery (15 tests): classification of every kind; downsample with placement invariance and `extractImages` size; gray → gray JPEG; soft-mask co-resampling with render PSNR; CMYK → RGB under `screen`, kept under `print`; CCITT output for bilevel with bit-exact renders; lossless changes no pixel (and rejects any undecodable row); unsupported kinds kept with a reason; shared XObject decided once; image inside a form; idempotence across saves and within one session. Sugar: `compress(images:)`. Stress: the 1000-page fixture.
- New form battery (`pdf_form_battery.dart`, 8th shared battery); every sweep test proves by read-back after save, render, or extraction; 7 fail-proofs per work package recorded in the hands' reports.
- `make check` green (analyze, native, all web lanes, example apps).

## Docs
Changelog bullets are one sentence each, enforced by `tool/check_changelog_shape.sh`. Roadmap, README (editor example, `compress`, capability table), ARCHITECTURE (source tree, upstream-patch table), MIGRATION (`imageQuality`/`imageScale` → `PdfImagePolicy`), example app (`Reduce images` op) and its smoke test, `CHANGELOG.pre.md` `5.0.0-dev.0`, `CHANGELOG.md` `5.0.0`.


lib/src/ops/pdf_editor.dart
```diff
@@ -301,6 +303,16 @@ class PdfEditor {
     );
   }
 
+  /// Lists the images placed on [page] (0-based): each one's resource
```
slopfairy, line 384: `pageImages` promises *image* XObjects, but the engine behind it returns *every* XObject drawn on the page. `get_page_images` (pdf_oxide `src/editor/document_editor.rs`) pushes an `ImageInfo` for each `Do` operator with no `/Resources` lookup or Subtype check — its own comment reads "For now, include all XObjects; a more refined implementation would check the XObject's Subtype." A page with Form XObjects (LaTeX, InDesign, office exports all produce these) lists them here too, and those names are not valid `resizeImage` targets, so the round-trip this PR advertises breaks on the first form. The dart-pdf fixture has no forms, which is why the battery is green.

Fix: filter to `Subtype /Image` in `edit_page_images` (dispatch.rs) before serializing, or correct the contract here and in `PdfPageImage` — rename to XObjects and drop the `resizeImage` guarantee. Filtering is what the roadmap row actually wants.

slopfairy (changes requested):
> [!CAUTION]
> Requesting changes — `pageImages` returns every XObject, not just image XObjects

1. `lib/src/ops/pdf_editor.dart:306` — `pageImages` documents a list of image XObjects whose names feed `resizeImage`, but the engine's `get_page_images` records every `Do` with no Subtype check, so Form XObjects are listed too and their names do not resize. Filter to `/Image` in the dispatch, or rename/document the API as XObjects.

✅ **Checked**
- Op name `editorPageImages` matches across the registry, `EngineOp`, the codec builder, and the shared-bridge handle method.
- `decodePageImages` field names line up with the Rust `put_map_list` (`name`, `x/y/width/height`, `a..f`); f32→f64 is lossless.
- Out-of-range and negative pages error via the guard in `get_page_images` — no panic path.
- Bounds/transform semantics match the docs: bounds are the transform translation plus scale magnitudes, transform carries the full geometry.
- Roadmap, README, ARCHITECTURE, and both changelogs are consistent with the change.

slopfairy (dismissed):
> [!TIP]
> Approved — image reducer, editor image ops, and the pageImages filter all verified against the engine.

1. `lib/src/ops/pdf_editor.dart:316` — the `pageImages` doc says the name "is what [resizeImage] takes"; it is also what `repositionImage` and `setImageBounds` take, so name all three, matching the shared vocabulary the PR advertises.

✅ **Checked**
- Prior finding fixed: `edit_page_images` now intersects upstream's `Do` walk with `page_image_xobject_names` (Subtype `/Image`, inherited `/Resources` walked), so forms no longer leak into `pageImages`. Resolved the open thread.
- Compose fix present: `reposition_image`/`set_image_bounds` write `wanted * inverse(current_ctm)`, not absolute values, so images under an enclosing transform move by the offset requested.
- Writer fix present: `write_full_to_writer`'s resource walk goes through `staged_or_source_object`, so staged replacements reach the output instead of being dropped.
- `policy.rs` decision table is pure and matches the Dart `PdfImagePolicy` docs: lossless never picks a lossy codec, CMYK→RGB only when asked, a JPEG is re-encoded only when downsampled, bilevel goes to CCITT G4.
- Wire codec round-trips: enum wire names in `report.rs` line up with `PdfImageEncoding`/`PdfImageColor`/`PdfImageAction`/`PdfImageKeepReason`, and `encodeImagePolicy`/`decodeImageReport` field keys match.
- Breaking changes documented (CHANGELOG 5.0.0, MIGRATION) and `reduceImages`/`compress` map to the `render` capability for the keep pipeline.

slopfairy (dismissed):
> [!TIP]
> Approved — the `pageImages` form-filter fix landed, and the Dart↔Rust wire contract checks out end to end.

1. `lib/src/types/pdf_image_policy.dart` — `jpegQuality` and `minPixels` are accepted without validation; the engine silently clamps them on the wire (`jpegQuality` to 1–100, `minPixels` to ≥0). Either document the clamp in the field docs or assert in the constructor, so a typo like `jpegQuality: 1000` fails loudly instead of quietly becoming 100.

✅ **Checked**
- Prior blocking finding is fixed in the engine: `edit_page_images` now takes the raw `Do` placements and filters them through the page's image XObject names (`.filter(|i| image_names.contains(&i.name))`), so Form XObjects no longer surface in `pageImages`. The comment "Upstream records every `Do` (forms included); the contract is images" confirms the fix targets exactly what was reported.
- FFI wire contract verified against the submodule at `f7e81fe`: request keys (`colorDpi`, `grayDpi`, `monoDpi`, `jpegQuality`, `allowLossy`, `convertCmykToRgb`, `minPixels`), response keys (`objectId`, `softMask`, `ppiMin` with the `-1` sentinel, `keepReason`, `bytesBefore`/`bytesAfter`, `widthAfter`/`heightAfter`), and every enum wire name (encoding, color, action, keep-reason) match the Dart enums in `pdf_image_report.dart` exactly.
- `pageImages`, `resizeImage`, `repositionImage`, and `setImageBounds` share one name vocabulary; the bounds + full six-coefficient transform decode is sound, and the new ops route through the shared bridge with no platform-specific code.
- `PdfImagePolicy` presets match Ghostscript's resolutions (`screen` 72/72/300, `ebook` 150/150/300, `print` 300/300/1200), and `reduceImages` is gated under the `render` capability alongside `compress`, consistent with the engine's cfg gates.
- No `.github` workflow changes in this PR.

slopfairy (dismissed):
> [!TIP]
> Approved — the one blocking finding from the prior round (pageImages listing Form XObjects) is fixed and now proven by a fixture and a battery test.

✅ **Checked**
- Prior finding resolved: `pageImages` now filters through the page's `/Resources /XObject` to `/Subtype /Image`. `test/fixtures/handwritten.dart` draws an image and a Form XObject on one page, and the editor battery asserts only the image is listed — the thread is closed.
- `lib/src/ops/pdf_editor.dart` — `pageImages`, `repositionImage`, `setImageBounds`, and `reduceImages` all call `_check()` before touching the handle and share the resource-name vocabulary the PR advertises.
- FFI boundary: `encodeImagePolicy`, `decodeImageReport`, and `decodePageImages` in `codec.dart` are consistent with the wire types in `binary_codec.dart`. Unknown enum names on the wire throw `PdfEngineError` via `_enumByWireName`, keeping engine failures a `PdfError` rather than a `StateError`.
- `reduceImages` and `compress` are both mapped to the `render` capability in `capabilities.dart`, matching the engine's render feature gate; the pure editor edits (`pageImages`, `resizeImage`, `repositionImage`, `setImageBounds`) correctly stay core, ungated.
- New types are exported from `lib/pdf_manipulator.dart`; `PdfImagePolicy` presets and `PdfImageReport` getters are coherent, and the `compress(images:)` sugar replacement reads clean.
- No `.github/workflows` files in the changeset, so there is no workflow-security surface to flag on this PR.

slopfairy (dismissed):
> [!TIP]
> Approved — the pageImages form-filtering finding is fixed with a regression test, and the Dart boundary for the new types is consistent end to end.

1. `lib/src/types/pdf_image_policy.dart:23` — range-check the numbers at the boundary. `jpegQuality` outside 1–100, a non-positive `threshold`, or `minSavings` above 1 all cross the wire unchecked; `minSavings: 1.5` can never save and a bad quality value fails far from the caller. A few asserts in the constructor surface programmer errors where they happen.

✅ **Checked**
- Prior finding: `test/ops/core/pdf_editor_battery.dart` now asserts `pageImages(0)` on the `imageAndFormPdf` fixture returns exactly `['Im1']`, so the Form XObject is filtered and the thread is resolved.
- `lib/src/bridge/protocol/codec.dart` — `encodeImagePolicy` omits `null` resolutions; `decodeImageReport` routes every wire enum through `_enumByWireName` (fails loudly on an unknown name); `decodePageImages` rebuilds bounds plus transform.
- `lib/src/bridge/protocol/binary_codec.dart` — the new fields ride existing tags (`_tI32`/`_tF64`/`_tString`/`_tMapList`); no new wire type to mismatch across platforms.
- `lib/src/ops/pdf_editor.dart` and `pdf_sugar.dart` — `pageImages`/`resizeImage`/`repositionImage`/`setImageBounds` share one name vocabulary; `compress` records `render` like `reduceImages`, matching the capability gate in `lib/src/keep/capabilities.dart`.
- Breaking changes and the `5.0.0` bump line up across CHANGELOG and MIGRATION.

The five-stage reducer itself lives in the `vendor/pdf_oxide` submodule, which I couldn't diff here (the local shell is unavailable in this sandbox), so the engine logic is covered by the test battery and fixtures rather than by direct reading.

slopfairy (dismissed):
> [!TIP]
> Approved — clean Dart surface and a properly regression-tested fix for the `pageImages` form leak; one stale README name to touch up

1. `README.md:349` — change `` `getSignatures` `` to `` `signatures` ``. The capability list still names the removed method (renamed to a getter in this PR), so a reader copying it from the README hits a compile error.

✅ **Checked**
- The prior `pageImages` finding is genuinely fixed, not just claimed: `test/fixtures/handwritten.dart` adds `imageAndFormPdf` (a `/Subtype /Image` XObject and a `/Subtype /Form` XObject on one page, resources inherited from `Pages`), and `pdf_editor_battery.dart:420` asserts `pageImages(0)` returns only `['Im1']`.
- FFI boundary round-trips: `encodeImagePolicy` omits null resolutions (engine reads "never downsample"), and `decodeImageReport` / `decodePageImages` decode every new field; enum wire names (`auto`/`full`/`half`, action and keep-reason names) line up on both sides.
- Capability mapping is right: `reduceImages` and `compress` both hit `KeepRecord.op('render')` and `PdfCapability.apiMembers`, so `keep: auto` trims the image pipeline correctly.
- Breaking renames are clean everywhere except the README spot above — no stale `getX()`, `optimizeImages`, or `imageQuality:`/`imageScale` references remain in Dart, examples, or tests (the `getSignatures` wire name in `op.dart` is correct and unchanged).
- `compress(images:)` defaults to `PdfImagePolicy.ebook` and still does a full rewrite with stream compression + GC, matching the changelog's "225 ppi" claim (150 × 1.5).

The bulk of the reducer and the transform-composition fix live in the `vendor/pdf_oxide` submodule pointer, which isn't reviewable from this repo's diff; the Dart side and the test battery that exercises it are what I could verify, and they hold up.

slopfairy (dismissed):
> [!TIP]
> Approved — the image reducer, image-only `pageImages`, and the reposition/setImageBounds edits are coherent, correctly wired through the bridge, and backed by a deep test battery.

✅ **Checked**
- Prior finding is fixed: `pageImages` now returns only `/Subtype /Image` XObjects. `test/fixtures/handwritten.dart` adds `imageAndFormPdf`, which draws `/Im1` and `/Fx1` on one page through inherited `/Resources`, and `pdf_editor_battery.dart` asserts only `Im1` is listed. The thread on this is resolved.
- The rename sweep (`getTitle()`→`title`, `getSignatures()`→`signatures`, `getPageMediaBox()`→`pageMediaBox()`, etc.) is consistent across `PdfEditor`, `PdfDoc`, the `BridgeEditorHandle`/`BridgeDocHandle` interfaces, and both shared-bridge implementations. The wire op names (`getSignatures`, `editorPageMediaBox`) are unchanged where the protocol needs them.
- `reduceImages` encode/decode round-trips: `encodeImagePolicy` omits null resolutions and sends `chroma` by enum name; `decodeImageReport` maps the `-1` ppi sentinel to null, resolves typed enums, and throws `PdfEngineError` on unknown wire names. All covered in `codec_test.dart`.
- The new wire fields reuse existing binary-codec types (f64, string, int, map-list); `pageImages` decoding and the report/policy paths are unit-tested at the codec layer.
- `lib/src/keep/capabilities.dart` maps `reduceImages`, `compress`, and `render` under the `render` cargo feature; the new edit ops (`pageImages`, `resizeImage`, `repositionImage`, `setImageBounds`) are core and correctly left un-gated.
- The editor battery covers the full reduce pipeline (classification, downsample placement invariance, soft-mask co-resampling with PSNR, CMYK→RGB under screen/kept under print, CCITT G4 bit-exact renders, lossless pixel identity, shared-XObject decide-once, image-in-form, idempotence) plus the position/bounds round-trips.

The Rust engine itself (vendor submodule) is not checked out in the workspace, so I reviewed the Dart surface and the test fixtures rather than the engine internals; the fixture and battery coverage exercises the engine paths end to end.

lib/src/bridge/protocol/codec.dart
```diff
@@ -321,6 +340,173 @@ List<SearchResult> decodeSearchResults(Map<String, Object?> r) {
...
+  'chroma': policy.chromaSubsampling.name,
+  'recompressJpeg': policy.recompressJpeg,
+};
+
+T _enumByWireName<T extends Enum>(List<T> values, Object? wire, String what) {
+  final name = wire as String? ?? '';
+  for (final v in values) {
+    if (v.name == name) return v;
+  }
+  throw PdfEngineError('unknown image $what "$name" on the wire');
```
slopfairy: `_enumByWireName` is shared by `decodeFormFields` (field type), `decodeImageReport` (encoding/color/action/keep reason), so the hardcoded `image` makes a form-field decode failure read "unknown image field type". Drop the word.

```suggestion
  throw PdfEngineError('unknown $what "$name" on the wire');
```

slopfairy (dismissed):
> [!TIP]
> Approved — the Dart surface, wire contract, and the image Decide table are all consistent; only a cosmetic codec nit remains

1. `lib/src/bridge/protocol/codec.dart:465` — `_enumByWireName` hardcodes the word `image` in its error string, so a form-field type decode failure reports "unknown image field type". Change `'unknown image $what ...'` to `'unknown $what ...'` (inline suggestion posted).

✅ **Checked**
- Breaking renames (`getTitle`→`title`, `getSignatures`→`signatures`, `getPageMediaBox`→`pageMediaBox`, `optimizeImages`→`reduceImages`, `compress(imageQuality:)`→`compress(images:)`) are applied consistently across `PdfEditor`, `PdfDoc`, `PdfStandalone`, `PdfSugar`, the `BridgeEditorHandle`/`BridgeDocHandle` interfaces, and `_SharedEditorHandle`/`_SharedDocHandle`. No stale `getXxx` left behind.
- Wire contract cross-checked against the engine: `pdf_oxide` `src/host/images/report.rs` emits `raw/jpeg/jpx/ccitt/jbig2/unknown`, `bilevel/gray/rgb/cmyk/other`, `kept/recompressed/downsampled`, and the 13 `keepReason` names — all byte-for-byte the Dart enum `.name` values, and the flat report fields (`objectId`, `ppiMin`, `softMask`, `bytesBefore/After`, …) match `decodeImageReport`.
- `src/host/images/policy.rs` Decide table is sound: `ppi_min` takes the largest (most demanding) placement, downsample triggers only past `target × threshold` and scales to the target, a stored JPEG is never re-encoded in place unless `recompressJpeg && allowLossy`, CMYK goes to JPEG only under `convertCmykToRgb` (else lossless Flate), and `allowLossy: false` never selects a lossy codec. All 10 rules are unit-tested and the tests assert the exact outcomes, not just pass/fail.
- New types all exported from `lib/pdf_manipulator.dart`; `PdfImagePolicy` presets' documented ppi/quality/chroma/CMYK numbers match their constructor args.
- Prior round finding is resolved: `pageImages` now filters through the page's `/Resources /XObject` to `/Subtype /Image` (the thread is marked resolved, and the PR body documents the fix + fixture).

One honest caveat: the engine lives in the `vendor/pdf_oxide` submodule (bumped `f12294c` → `ef6d1ccb`), which is not checked out in this workspace, so I reviewed the Dart side in full and spot-checked the highest-risk Rust modules (`policy.rs`, `report.rs`, the commit diffs) rather than every line of the ~90 KB image pipeline. The 20 Rust unit tests plus the 15-test editor battery and the three-producer fixtures are what carry the remaining confidence there.

slopfairy (approved):
> [!TIP]
> Approved — the one open thread from last round is fixed and the full changeset is clean.

The only new commit since my last review is the codec error-name fix, and it does exactly what the thread asked.

✅ **Checked**
- `lib/src/bridge/protocol/codec.dart` — `_enumByWireName` now takes a `what` and throws `unknown $what "$name"`, so a form-field decode failure no longer reads "unknown image field type". Thread resolved.
- Breaking surface matches the PR body: noun getters (`title`, `author`, `subject`, `keywords`, `producer`, `creationDate`, `signatures`) and noun methods (`pageMediaBox(page)`, `pageImages(page)`), `reduceImages(PdfImagePolicy)`, `compress(images:)`, `applyRedactions()` returning `PdfRedactionReport`.
- New types are typed and documented with no raw PDF syntax leaking to callers: `PdfImagePolicy` (presets + ppi/quality/chroma/minSavings knobs), `PdfFormField` with sealed `PdfFormFieldValue` and `PdfFormFieldFlag` at the correct ISO `/Ff` bit positions, `PdfImageReport`/`PdfImageOutcome`, `PdfPageImage`, `PdfMatrix`, `PdfSanitizeOptions`, `PdfXfaInfo`, `PdfAttachment`.
- Layering rule holds throughout: reads on `PdfDoc`, mutations on `PdfEditor`, exports on `PdfDoc`/`PdfStandalone`. The bridge wiring (`shared_bridge_editor.dart`, `shared_bridge_doc.dart`, `op.dart`) is encode → `_exec` → decode with no stray logic.
- `binary_codec.dart` i64 split is correct: arithmetic high/low halves (not 32-bit bit masks) write the same bytes on VM and dart2js, and the comment bounds it to 2^53 which every count on this wire respects.
- Tool gates are sound: `check_changelog_shape.sh` strips link targets before counting words, `check_fork_tests.sh` compares against the base tag, and `run_web_test.sh` now kills chromedriver's forked browser and fails a hung drive at the deadline instead of blocking CI.

